PR #70, opened 2014-03-16
Add watch functionality for broccoli-cli.

garetht:
This is meant to close joliss/broccoli#58.


joliss:
I don't want to add watch right now (there's [broccoli-timepiece](https://github.com/rwjblue/broccoli-timepiece) for those who need it), but would be happy to revisit after Broccoli 1.0

Issue #144, opened 2014-05-13
Broccoli watcher not firing change event


jeffjo:
I am instantiating and using a Watcher object in order to regenerate build output on file changes:

``` javascript
var tree = broccoli.loadBrocfile();
var builder = new broccoli.Builder(tree);
var Watcher = require('broccoli/lib/watcher');
var watcher = new Watcher(builder, { verbose: true });

watcher.on('change', function(tree) {
  //copy tree.directory to some output directory
});
```

Every once in a while, the watcher does not seem to be able to detect changes to the file system and the 'change' event is never fired. At this point, the only thing I can do is restart the script. This happens pretty infrequently, but it seems to happen more often when I'm away from my computer for a while.

Can anyone give me any pointers on how to debug this issue? I've tried sending `SIGUSR1` to the node process, but when I start the node debugger, it doesn't respond to any of the debugging commands.

stefanpenner:
This problem hasn't come up for me in ages. New versions of node, new versions os's etc seem to have addressed this

Issue #72, opened 2014-03-17
Refactor Watcher to not poll?


ericf:
I've been noticing that the Broccoli Watcher task is taking up a constant 4% CPU on a medium-sized project. I think Watcher should _not_ poll by default, but instead preform a check every time its `then()` method is called.

If someone wants the Watcher to poll, then they could easily setup a `setInterval(watcher.check, 100)`. This would also ensure that when the middleware/server receives a request for a file, it will always be built and up to date.

@joliss Thoughts? If this seems good to you, I'd be happy to issue a PR for this change.

stefanpenner:
after long wait, broccoli ships with broccoli-sane-watcher by default. (WHich defaults to a non polling watcher).

PR #150, opened 2014-05-20
Integrate Chokidar into Watcher module

rondale-sc:
Integrate Chokidar into Watcher module instead of using setInterval.  I've smoke tested this against [broccoli-sample-app](https://github.com/broccolijs/broccoli-sample-app) and it has worked beautifully.

This resolves the sleep issue described in #144 and the CPU issue described in #72

Let me know what you think.  :))))


stefanpenner:
broccoli-sane-watcher has been rolled into 1.0.

PR #153, opened 2014-05-24
Call `read` only on the trees that have changed dependents

ppcano:
It removes the need to rerun all the filters when a file has changed, therefore it improves the build total time in some specific cases.

Optionally, the filter must define an `isUnchanged` function which defines if its output has changed since the last build.

``` js
MyFilter.prototype.isUnchanged = function () {

  var lastHash = this.inputFiles.hash,
      baseDir = this.inputFiles.baseDir,
      files = this.inputFiles.files;

  var hash = helpers.hashStrings( getFileKeys(baseDir, files) );
  return lastHash === hash;
}
```

Filters which depends on other filters will always provide its result based on the result of every dependent filter. 

This is a first version pending of review which has been successfully tested by implementing the `isUnchanged` function to the `StaticCompilerFilter`. This filter implementation is available at [static-compiler/is-unchanged](https://github.com/ppcano/broccoli-static-compiler/tree/is-unchanged) and [static-compiler/is-unchanged-with-symlink](https://github.com/ppcano/broccoli-static-compiler/tree/is-unchanged-with-symlink)


stefanpenner:
This is fantastic direction. Sorry this hasn't gotten the feedback it deserved. 

Since then, we have taken a different direction. @rwjblue and myself have been pairing on a solution to the same problem. What we came up with is the following:

1. annotate plugins as "side-affect free"
2. they invoke `this.revised()` if they have changed, they skip if they have not
3. if a plugin who has become "side-affect free"'s inputs have not changed, we will skip invoking its `build` method
4. for roots which are watched, we are tying them into the watcher, and marking them as `revised()` only if we observed a change from them, if a root isn't watch we must assume they are always `resived()`. We could also create an `immutable root` or sorts, for input that we know absolutely will not change.

Some of our work is being tracked: https://github.com/broccolijs/broccoli-plugin/pull/23
Most of our work still needs to be pulled out of our local spikes.

I am closing this, not because this direction isn't great, but because the code-base has diverged, and we have continued what I believe to be a similar effort elsewhere.

PR #152, opened 2014-05-31
Better build failure error handling

wagenet:
The current "Build failed" handling can hide relevant error information:

https://github.com/broccolijs/broccoli/blob/7e5048dec8823f937f0d59f57478f5f50559c6c8/lib/cli.js#L53-L59

For instance, UglifyJS has detailed information about parse errors that doesn't get shown by Broccoli. 

We could call `toString` on the error, but in this case we'd get a double stack trace since UglifyJS adds the trace to `toString`. I wonder if we should hide the trace by default and then allow for a `--trace` flag. This is common practice in some other libraries.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -50,9 +50,15 @@ function broccoliCLI () {
           process.exit(0)
         })
         .catch(function (err) {
-          // Should show file and line/col if present
+          if (err.message) {
+            console.error(err.message)
+          }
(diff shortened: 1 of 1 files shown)


stefanpenner:
I absolutely agree this should be improved. ember-cli's has been, and we are planning to role that work back in here.

The code has all changed since this was opened, so closing this for now.

PR #163, opened 2014-06-06
Replace watcher.js with broccoli-sane-watcher.

### LINKED ISSUE #72: Refactor Watcher to not poll?

jschilli:
Replaces watcher.js with @krisselden's broccoli-sane-watcher 

Should resolve #72 and replace #150

On our largish ember.js project we went from consistent 25% cpu with current watcher to roughly 0% using the sane-watcher

cc: @stefanpenner @rondale-sc @krisselden


krisselden:
@joliss should this be closed? You have a different plan for integration of broccoli-sane-watcher, confirm?

stefanpenner:
@krisselden ya i believe she does.

PR #171, opened 2014-06-17
Watch

jnicklas:
A simple solution to #70. It simply uses the watcher and updates a symlink whenever the target has changed. This seems to work very well for me. It also complains if the target already exists and isn't a symlink. I think this is a good solution because it is safe (accidentally overwriting a symlink isn't that bad), it's simple, and it's performant (no unnecessary copying).

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -29,6 +29,10 @@ function broccoliCLI () {
       actionPerformed = true
       var watcher = new broccoli.Watcher(getBuilder(), {verbose: true})
       watcher.on("change", function(path) {
+        if(!fs.lstatSync(outputDir).isSymbolicLink()) {
+          console.error('Error: "' + outputDir + '" already exists and is not a symbolic link. Refusing to overwrite files.')
+          process.exit(1)
+        }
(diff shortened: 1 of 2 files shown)


joliss:
Hm. I'm a bit on the fence on this one, but I'd prefer not to have a `watch` command in Broccoli core for now. [broccoli-timepiece](https://github.com/rwjblue/broccoli-timepiece) seems to adequately provides this kind of watch functionality.

I'll leave this PR open though in case I change my mind at some point.

greghuc:
@markuskobler, @joliss I've just updated dependencies in broccoli-timepiece, including broccoli 0.13.3. See this pull request: https://github.com/rwjblue/broccoli-timepiece/pull/10. Would be great if @rwjblue merged this soon.

I've been happy using broccoli-timepiece as a watch command. However, it has lagged on bug-fixes. I've been seeing double builds for ages: apparently this was fixed in broccoli 0.12.0!

joliss:
I don't want to add this right now, but we can perhaps revisit this in the future, after Broccoli 1.0.

Issue #192, opened 2014-09-16
encourage common logging patterns


stefanpenner:
As most of the complexity lies in various filters, it might be a good idea for us to have a suggested set of logging conventions that we encourage plugin/filter authors to abide to.

That way advanced logging can be enabled globally for all broccoli filters, and hopefully make various debugging situations nicer.

stefanpenner:
We now recommend: https://github.com/heimdalljs/heimdalljs-logger

- [ ] update readme

PR #194, opened 2014-10-04
Support for .brocfileLocation in error objects

diogoko:
Adds support for a `.brocfileLocation` property on error objects thrown by plugins as suggested by @rwjblue in issue #192.

Setting the property in `broccoli-writer` was not possible as not every plugin calls its parent's constructor (eg: `broccoli-es6-concatenator`), but decorating the error is feasible with a simple:

```javascript
    }).catch(function(err) {
      if (self.brocfileLocation) {
        err.brocfileLocation = self.brocfileLocation
      }
      throw err
    })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,14 @@ easier to track down by setting the `.file` property on the generated error.
 This `.file` property is used by both the console logging, and the server middleware
 to display more helpful error messages.
 
+Due to the asynchronous architecture of Broccoli, stack traces may not reflect the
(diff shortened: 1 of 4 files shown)


joliss:
I'm closing this for now. I'd love to have stack traces pointing at where a plugin was instantiated in Brocfile, but we'd need a solution that doesn't require adding code to every plugin. I'm not sure such a solution exists, with the current tooling.

PR #208, opened 2014-11-07
Add CORS header

jnicklas:
When developing with Broccoli, sometimes one might want to serve up font files from the Broccoli server. However, certain browsers restrict access to font files based on CORS rules. This makes it impossible to have broccoli serve up font files to another application running on a separate port. This simply adds a CORS header which allows access to these files.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -1,5 +1,6 @@
 var path = require('path')
 var fs = require('fs')
+var dns = require('dns')
 
 var handlebars = require('handlebars')
 var url = require('url')
@@ -99,14 +100,26 @@ module.exports = function getMiddleware(watcher, options) {
       response.setHeader('Cache-Control', 'private, max-age=0, must-revalidate')
       response.setHeader('Content-Length', stat.size)
       response.setHeader('Content-Type', type)
-      // add CORS headers for font files
-      response.setHeader('Access-Control-Allow-Origin', '*')
 
-      // read file sync so we don't hold open the file creating a race with
-      // the builder (Windows does not allow us to delete while the file is open).
-      buffer = fs.readFileSync(filename)
-      response.writeHead(200)
-      response.end(buffer)
+      function writeResponse() {
+        // read file sync so we don't hold open the file creating a race with
+        // the builder (Windows does not allow us to delete while the file is open).
+        buffer = fs.readFileSync(filename)
+        response.writeHead(200)
+        response.end(buffer)
+      }
+
+      if(request.headers.origin) {
+        // Check if origin resolves to a loopback IP, only add CORS headers if that's the case.
+        dns.lookup(url.parse(request.headers.origin).hostname, function(err, address) {
+          if(!err && address === '127.0.0.1') {
+            response.setHeader('Access-Control-Allow-Origin', request.headers.origin)
+            writeResponse()
+          }
+        });
+      } else {
+        writeResponse()
+      }
     }, function(buildError) {
       var context = {
         message: buildError.message || buildError,
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -99,6 +99,8 @@ module.exports = function getMiddleware(watcher, options) {
       response.setHeader('Cache-Control', 'private, max-age=0, must-revalidate')
       response.setHeader('Content-Length', stat.size)
       response.setHeader('Content-Type', type)
+      // add CORS headers for font files
+      response.setHeader('Access-Control-Allow-Origin', '*')
 
       // read file sync so we don't hold open the file creating a race with
       // the builder (Windows does not allow us to delete while the file is open).


jnicklas:
So dredging this up, since I am still running an `npm link`'ed broccoli on all machines in our office to get this silly fix in.

To address @stefanpenner's concerns:

1) Production and development inherently differ. Anyone who is serving assets statically in production is going to run into this sooner or later. That's what staging environments are for. "Oh, the fonts are not working, wonder why that is?", *opens console*, "Aha!" / "Wtf is CORS?". If this actually causes a serious problem for someone in production they have only themselves to blame.

2) Even if someone somehow broke the code to add the CORS header to requests to the hypothetical proxy server, it *still* wouldn't be a security issue. There are two kinds of CORS requests, simple and preflighted. Preflighted requests require an additional OPTION request before the action, and the server must provide `Access-Control-Allow-Methods` to specify which methods are permitted. Since we do not specify this header, a preflight request will never succeed.

Simple requests, which are basically GET and POST with certain content encodings, do not need preflight requests. It is only these requests that this PR permits. Why are the rules more relaxed for these requests? Because these can be performed without the use of XMLHttpRequest already, through cross origin form posts and cross origin resource loading respectively, which all browsers already permit. This is why we have CSRF protection in web frameworks. In short, enabling CORS on these kinds of requests doesn't expose any attack vector which can't already be exploited in other ways.

Let's not forget that CSRF is a session attack. If the app in question does not make any use of session data, and the broccoli server does not, then it cannot be exploited.

Personally I find it silly to make this optional. It just makes things needlessly frustrating for users. That being said, if after this you still feel the need to make this an option I will admit defeat and amend the PR. Your choice.

stefanpenner:
> So dredging this up, since I am still running an npm link'ed broccoli on all machines in our office to get this fix in.

Ya this is a problem, it should be configurable or some possible for you to get what you want with the official release. Having to link the project is a deficiency, and should be addressed.

@joliss your feedback would be great, as my above comments are mostly speculative assumptions of your opinion.

joliss:
It seems that if we always add `Access-Control-Allow-Origin: *` as suggested by this PR, we make Broccoli users vulnerable to having all their file contents siphoned: For instance, an attacker might create an evil page http://wateringhole/, which performs lots of XHR requests to http://localhost:4200/... and sends the file contents back to http://wateringhole/.

Side note: It's true that we already allow cross-origin embedding e.g. via `<script>`, but this is much harder to exploit. I'd love to mitigate this, but [MDN says](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#How_to_block_cross-origin_access) that the only way to stop cross-origin embedding is to use a CSRF token. (I'm very surprised. Is this really true?) So pragmatically there seems to be nothing we can do about this.

Anyways, back to getting fonts working: We might add an option for a CORS header, but I'm really hesitant to add options to the core Broccoli package, especially pre-1.0. Perhaps we'll have a separately-versioned server package some time; then it's easier to add these things.

I don't suppose there is a way to make CORS work for fonts without introducing security issues, is there? (Extension-sniffing doesn't count.)

In the absence of a nice optionless fix, can we revisit this in a few months?

jnicklas:
Sorry to be taking so long to get back to this.

@joliss that's a really good point.

I've updated this PR to a more conservative option, where we're adding an access-control-allow-origin header only if the origin domain resolves to `127.0.0.1`. The code isn't super pretty, but I'm not sure how to make it read better given that `dns.lookup` is async. Given that the attacker cannot influence the origin header, and `wateringhole.com` resolves to some other IP, this should prevent this attack.

Unfortunately this isn't an ideal solution since it doesn't allow fonts to be loaded from inside a VM for example, or anywhere else where we want to access them through an actual IP rather than a loopback IP.

devongovett:
:+1: on this. Came across this problem for both fonts and web worker scripts today. Any updates?

stefanpenner:
> web worker scripts today

I believe CORs headers won't help you as WebWorker spec has no awaresness of CORs. I reported this as a bug quite some time ago, but I don't believe any progress has been made.

context: http://www.w3.org/TR/workers/#dom-workerglobalscope-location
SO: http://stackoverflow.com/questions/20410119/cross-domain-web-worker

A work-arounds for workers is host them local, or consider the risky eval running worker. That accepts scripts as input. This is dubious at best but until the spec/vendors/browsers fix ^. I am uncertain of a better solution :(

@devongovett but @joliss's comment: https://github.com/broccolijs/broccoli/pull/208#issuecomment-69260724 is still the status of this issue.

devongovett:
I worked around the Worker issue by using `importScripts` from within a worker created from a blob url, which oddly enough doesn't seem to enforce any sort of same-domain requirement.

```
var blob = new Blob(['importScripts("http://localhost:4200/worker.js");'], { type: 'text/javascript' });
var worker = new Worker(URL.createObjectURL(blob));
```

Can't seem to get around the font issue though.

stefanpenner:
> blob url,

ah yes, blob URLs.

I think when I attempted, i was just pwnd bug buggy behavior on old Android and lack of support on now "aged out" versions safari.

jnicklas:
@stefanpenner I have tried to address @joliss's concerns. The current iteration of this PR doesn't solve the issue completely, but it improves on the status quo. It should not be vulnerable to the sniffing method that @joliss was concerned about and it does not introduce a new option.

lib/middleware.js
```diff
@@ -86,11 +87,25 @@ module.exports = function(watcher) {
...
+        // read file sync so we don't hold open the file creating a race with
+        // the builder (Windows does not allow us to delete while the file is open).
+        buffer = fs.readFileSync(filename)
+        response.writeHead(200)
+        response.end(buffer)
+      }
+
+      if(request.headers.origin) {
+        // Check if origin resolves to a loopback IP, only add CORS headers if that's the case.
+        dns.lookup(url.parse(request.headers.origin).hostname, function(err, address) {
```
joliss: Do we *need* a DNS lookup here, or do you think we get away with comparing to `localhost`, `127.0.0.1`, `::1` for performance?

joliss:
> Unfortunately this isn't an ideal solution since it doesn't allow fonts to be loaded from inside a VM for example, or anywhere else where we want to access them through an actual IP rather than a loopback IP.

I agree - it might be an acceptable partial solution though. I'm kind of on the fence.

Is what we're doing in this PR safe? I think it probably is, but I'm having trouble convincing myself that it's definitely safe.

@stefanpenner, @rwjblue, do you have an opinion on all this?

stefanpenner:
If CORs must exist, my vote is for it to be an opt-in.

lib/middleware.js
jnicklas: The DNS lookup is convenient for people who run their dev servers on a local domain name (think /etc/hosts). Something like [pow](http://pow.cx/) would work with this DNS lookup, whereas a simple string comparison wouldn't work. Maybe that's acceptable though, it would definitely make for prettier code.

stefanpenner:
I believe we should address this via @jakedetels "enhanced customization of server options" https://github.com/broccolijs/broccoli/pull/292

lifeart:
related: https://github.com/broccolijs/broccoli/pull/490

PR #209, opened 2014-11-09
dist folder is now default if build folder is omitted.

Globegitter:
I quite got to like this from ember-cli.


joliss:
I'm not sure... I think I'd prefer not to merge this. It's great for an opinionated tool like Ember CLI, but plumbing libraries like Broccoli should be explicit.

PR #212, opened 2014-11-15
Use the connect-history-api-fallback middleware to redirect 404s to /

eploko:
This adds a simple connect middleware to the current server implementation to fallback SPA browser reloads to the root path.

I've skimmed through the currently open PRs and issues and it's not clear if it's ever planned to support this in Broccoli. Please let me know if it's a wrong way to go.

P.S. Broccoli rocks. :)

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -3,6 +3,7 @@ var middleware = require('./middleware')
 var http = require('http')
 var tinylr = require('tiny-lr')
 var connect = require('connect')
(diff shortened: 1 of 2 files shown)


joliss:
I don't want to support this in Broccoli core at the moment. I think tools built on top of Broccoli, like Ember CLI or Cabbage, are in a better position to decide how they want to serve stuff.

PR #218, opened 2014-12-03
Fail fast if outputDir already exists, instead of building first

### LINKED ISSUE #217: broccoli build command should fail-fast if outputDir already exists
If the `<target>` dir already exists, `broccoli build` should fail with an error message before doing the build first.

naartjie:
Fixes #217


joliss:
Landed in 073fed4; thanks!

Issue #202, opened 2014-10-09
Proposal: Centralize Temp Directory Assignment


rwjblue:
Currently, each plugin is choosing its own temp directory (generally hard-coding to `./tmp` in [quick-temp here](https://github.com/joliss/node-quick-temp/blob/6799b270edb202144747bb22b8cfe85dd5bf9df4/index.js#L40-L54)).  We need to be able to centralize this location, to allow usage of the system temp directory (or whatever option is passed).

I see a couple of possible ways forward (in order of my personal preference):

1. Set a property named `tmpRoot` on each tree before calling the `read` function.  This would still easily allow utilities like `quick-temp` or the plugins themselves to handle the generation / deletion of actual directories to be used, but allows Broccoli to control the root location of the temp dirs.

2. Add logic to the Broccoli builder itself to create and manage the temp directories for each tree providing the generated paths to the tree itself.  This allows Broccoli to manager the exact location of each temp directory.

3. Set `process.env.BROCCOLI_TMP_ROOT` inside the builder, and let all plugins read that environment variable.

4. ????

raytiley:
@rwjblue @joliss would you two be able to do a hangout this weekend or next week sometime? Would love to nail down a plan of action on this and get to work on it :)

raytiley:
@rwjblue @joliss ping.

joliss:
The new [`.rebuild` API](https://github.com/broccolijs/broccoli/blob/master/docs/new-rebuild-api.md) will centralize all control over tmp dir handling in the Broccoli builder. Once support for old-style plugins is dropped in Broccoli core, this will become configurable when instantiating the builder.

stefanpenner:
woot!

Chinow:
Hi,

Is this still planned (.rebuild API link is 404)?
I have to work with Visual Studio and TFS and they keep monitoring file changes in every single folder even those in the .tfignore file.
Moving away the tmp folder wil prevent insane freeze and crashes.

joliss:
Yes, still very much planned!

stefanpenner:
I suspect the patched based rebuilds will also dramatically cut down on issues. As the majority of write io will vanish

PR #224, opened 2014-12-12
Set tmpRoot on tree if tmpRoot provided to builder

raytiley:
I know this needs tests, but I wanted to start the discussion. This is for #202.


stefanpenner:
broccoli 1.x now supports configurable tempdirs.

PR #225, opened 2014-12-19
Add a force option on build.

bruce-one:
Adds a --force (or -f) option on building which will remove the target if it's a pre-existing directory. Hence making the "EEXIST" error avoidable if desired.

Fixes [broccoli-cli issue 4](https://github.com/broccolijs/broccoli-cli/issues/4)

This adds shelljs as a dependency, which is a bit of a strange library - so I understand if you're not keen on it :-) (I'd probably, personally, expect use fs-extra for this task - but don't mind, and shelljs seemed interesting :-) (and is already used by a broccoli dependency :-) )

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,7 +1,7 @@
 var fs = require('fs')
 var program = require('commander')
 var copyDereferenceSync = require('copy-dereference').sync
-
+var shelljs = require('shelljs')
 var broccoli = require('./index')
 
 module.exports = broccoliCLI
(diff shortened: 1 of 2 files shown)


joliss:
Thanks for the PR. I've gotten a request for this several times, but I don't want this in Broccoli core, because it's too easy to accidentally cause data loss by pointing it at the wrong directory.

There'll usually be some tool in charge of a project's directory hierarchy (e.g. ember-cli, a Gruntfile, etc.), and that tool is better equipped to delete stuff, because it knows which directories are safe to clear.

PR #239, opened 2015-03-20
* hack for live-reload

stefanpenner:
* ansi-colored message

### DIFF
--- a/templates/error.html
+++ b/templates/error.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <title>Build error</title>
+    <script src="/ember-cli-live-reload.js" type="text/javascript"></script>
   </head>
   <body>
     <h1>Build error</h1>
@@ -11,7 +12,7 @@
         {{#if treeDir}}<span style="color: #888">(in <code>{{treeDir}}</code>)</span>{{/if}}
       </p>
     {{/if}}
-    <p><b><code>{{message}}</code></b></p>
+    <p><b><pre>{{{message}}}</pre></b></p>
     {{#if stack}}
       <hr>
       <pre style="color: #888">{{stack}}</pre>
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -1,6 +1,8 @@
 var path = require('path')
 var fs = require('fs')
-
+var ansi2html = require('ansi2html')
+var ansi_up = require('ansi_up');
+ 
 var handlebars = require('handlebars')
 var url = require('url')
 var mime = require('mime')
@@ -93,7 +95,9 @@ module.exports = function(watcher) {
       response.end(buffer)
     }, function(buildError) {
       var context = {
-        message: buildError.message || buildError,
+        message: buildError.message &&  ansi_up.ansi_to_html(buildError.message, {
+          use_classes: true
+        }) || 'Build Error',
         file: buildError.file,
         treeDir: buildError.treeDir,
         line: buildError.line,


templates/error.html
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <title>Build error</title>
+    <script src="/ember-cli-live-reload.js" type="text/javascript"></script>
```
stefanpenner: this is a hack, but a simple way to get live-reload from the containing server.

We could make this configurable, or.... maybe even better just make the whole template configurable. This would allow easy customization

timmfin: So the purpose of this is to get live-reload without requiring a browser extension?

stefanpenner:
Ya ember-cli just gives you live reload. Without you need inc the browser extension. Unfortunately when we live reload you to the error screen, we don't live reload any longer.

templates/error.html
joliss: Adding some customization hook for ember-cli to load scripts here sounds great. I'd like to land the `.rebuild` API, and then redo the builder, watcher, and server on top. Is it OK to wait till then?

joliss:
As for coloring, I understand the use case, but I'm not a huge fan of ANSI escape sequences - I've used them a lot for prettifying things in my terminal, and every time they've been a pain in the neck. They also seem weirdly underdocumented with slightly different behavior for different terminals. Using ANSI sequences for terminal output seems OK out of necessity, but I'd prefer not to use them as a colorization "protocol" for htmlized browser output.

We could in principle add more error-reporting API, e.g. `message` = plain text, `ansi_message` = for terminal, `html_message` = for browser. But perhaps that's overkill.

I think my biggest question right now is: How much do we actually gain from coloring in practice? Do you have screenshots to share?

templates/error.html
stefanpenner: I believe the perfect solution, that would prevent polluting broccoli core, would be:

* ability to swap out `dir` + `error` templates
* ability to register handlebars helpers

Quirks of babel errors, they need to be cleaned-up for our use-case, as they include the nice error message in `error.message` and `error.stack`. They also always include the ansi-color sequences if the terminal supports them. Which is nice for the errors we log to the console, but not so nice for the web-browser.

stefanpenner:
> As for coloring, I understand the use case, but I'm not a huge fan of ANSI escape sequences - I've used them a lot for prettifying things in my terminal, and every time they've been a pain in the neck. They also seem weirdly underdocumented with slightly different behavior for different terminals. Using ANSI sequences for terminal output seems OK out of necessity, but I'd prefer not to use them as a colorization "protocol" for htmlized browser output.
> 
> We could in principle add more error-reporting API, e.g. message = plain text, ansi_message = for terminal, html_message = for browser. But perhaps that's overkill.
> 
> I think my biggest question right now is: How much do we actually gain from coloring in practice? Do you have screenshots to share?

All the coloring comes from babel, I don't really want/need the coloring currently, I am just using the ansi module to translate the ansi color codes -> HTML elements and classes. This allows future theming.

stefanpenner:
with this patch, and https://github.com/ember-cli/ember-cli/pull/3576 the output looks like:
 
* un-colored (as we transformed ansi color-codes -> html + classes, but left styles out for now)
* readable

joliss:
> All the coloring comes from babel

So babel is putting ANSI codes in their `err.message` properties? I'd argue that that is pretty suboptimal and needs to be fixed on their end. Not every `err.message` ends up in stdout/stderr, and not every stdout/stderr ends up in a terminal (e.g. `broccoli build &> debug.log`). Similar issue on CoffeeScript: https://github.com/jashkenas/coffeescript/issues/3407

Also, completely agree with the issue you raise re UI background.

sebmck:
> So babel is putting ANSI codes in their `err.message` properties? I'd argue that that is pretty suboptimal and should perhaps be changed on their end.

I agree which is why it's being changed :smile:

templates/error.html
joliss: Forking the middleware sounds good, and then we pull stuff back in once things have stabilized in Broccoli core.

stefanpenner:
> So babel is putting ANSI codes in their err.message properties? I'd argue that that is pretty suboptimal and should perhaps be changed on their end.
> I agree which is why it's being changed

what is the planned change

stefanpenner:
@joliss to get back on track, are you open to configurable templates and custom handlebar helper registrations? If so, we can do all of this from the outside world.

joliss:
@stefanpenner Probably yes, assuming the code doesn't look outright terrible... Keep in mind, anything you send to Broccoli right now wouldn't make it into ember-cli for, because ember-cli is stuck-ish on 0.13.6 until the API change is done. So we should probably hold off on tackling this in Broccoli core.

joliss:
Superseded by #243.

stefanpenner:
:+1:

PR #248, opened 2015-04-01
Use broccoli-read-compat for .rebuild tree wrapping.

rwjblue:
If a tree has a `.rebuild` function (and is not already wrapped in a compat layer), automatically wrap it with `broccoli-read-compat`'s `wrapInstance` method.


rwjblue:
I have tested this with a non-compat mode plugin (with only a `.rebuild` function), and have tested in ember-cli.

@joliss - Should be ready for review...

Issue #240, opened 2015-03-20
Clean up error API along with `.rebuild` change


joliss:
Should we clean up the error-reporting API for Broccoli plugins in the same go as the upcoming [`.rebuild` change](https://github.com/broccolijs/broccoli/blob/master/docs/new-rebuild-api.md)?

At the moment we have an informally-specified [set of error properties](https://github.com/broccolijs/broccoli/blob/affae661d64a5941e0c71d1214b4b46088890096/lib/middleware.js#L96-L100).

Just a reminder for anyone popping in, we're using these properties to separate out file/line/message/stacktrace info on structured error pages like this one:

![build_error](https://cloud.githubusercontent.com/assets/524783/6756428/d9668a5e-cf21-11e4-955e-1844f84e50dd.png)

In the future, we might even automatically show a few lines of file context, similar to how Rails does it.

Anyways, the current set of error properties is pretty ad-hoc and not very well thought out. Some changes we might want to make, off the top of my head:

* The `.message` property that Broccoli uses coincides with the JavaScript `Error.prototype.message` standard property. This is usually OK, but perhaps it's better to separate the two?
* Move everything into a `err.broccoliInfo` sub-object or so, to avoid polluting the main error namespace?
* Plugins that have multiple input trees may(?) need to communicate which tree an offending file came from. At the moment we have `.treeDir` (the base directory of the input tree) and `.file` (the relative path of the offending input file), where `path.join(err.treeDir, err.file)` is supposed to be the full file path. It's not clear to me whether this is even necessary. It seems a bit cumbersome.
* I remember seeing compilers with column ranges (like `.columnStart`, `.columnStop`). Perhaps we should support that.

Some of these changes might be breaking - so if we decide to do them, we should bundle them with the new `.rebuild` API.

Cc @iamstef, @rwjblue, and anyone else interested in plugin error reporting.

joliss:
Some changes we might want to make, off the top of my head:

* I remember seeing compilers with column ranges (like `.columnStart`, `.columnStop`). Perhaps we should support that.
* The `.message` property that Broccoli uses coincides with the JavaScript `Error.prototype.message` standard property. This is usually OK, but perhaps it's better to separate the two?
* Move everything into a `err.broccoliInfo` sub-object or so, to avoid polluting the main error namespace?
[...]

PR #249, opened 2015-04-03
add support for babel 5.0’s error.codeFrame

stefanpenner:


rwjblue:
Can you describe what a `codeFrame` is?  Maybe an example of the output with this?

stefanpenner:
> Can you describe what a codeFrame is? Maybe an example of the output with this?

@sebmck moved the code snippet of the source of the parse error out of the error + stack, it now just lives as .codeFrame

joliss:
@rwjblue It's a file context, like [this screenshot](https://github.com/broccolijs/broccoli/issues/240#issuecomment-84077683).

I'm actually wondering though, is there ever a case where we can't auto-generate the complete code frame from file/line/col? Babel's code frame doesn't seem to provide any info we don't already have in file/line/col. So rather than adding redundant information to the error object, perhaps we can add auto-generated code frames to the error page, and see if that might already cover all of our use cases...

stefanpenner:
> Babel's code frame doesn't seem to provide any info we don't already have in file/line/col. So rather than adding redundant information to the error object, perhaps we can add auto-generated code frames to the error page, and see if that might already cover all of our use cases...

I don't believe we currently have immediate access to the source babel choked on, so the line/col information will have to also rely on us threading the file content forward, or making the assumption that the path/file correlate to input file, and not some derived file.

At this point I don't really have the bandwidth to do the re-construction on our side.

rwjblue:
Many transformations may have happened from the file we know on disk. So the file, line, column info doesn't seem useful directly.

stefanpenner:
@joliss i can gladly append the codeFrame back to .message for the time being. Let me know.

joliss:
> At this point I don't really have the bandwidth to do the re-construction on our side.

That's fine, we can do it later.

I'm a bit hesitant to add API until we're actually convinced that we need it. So for the Babel case, let's tack it onto `.message` for now like you said, Stef, and once Broccoli auto-generates the code frame we can take it out again.

stefanpenner:
ember-cli 0.2.3 will see this in the stack.

Issue #7, opened 2014-01-12
Broccoli and express.js


MajorBreakfast:
I noticed a while ago that my projects get more and more top level files. `Gruntfile.js`, `.jshintrc`, `.gitignore`, `bower.json`, `package.json`, ... . All those files have their purpose. I would like Broccoli to work without a top level file. And here's how:

``` Javascript
// server.js

var app = require('express')();
app.use(app.router);
app.use(require('broccoli')(/* Stuff from Broccolifile.js */));

app.listen(8000);
```

The clou: Broccoli as an express middleware (i.e. `function(req, res, next) {...}`). It handles all the requests not handled by the router. If it's currently rebuilding it simply waits until it's done.

This solution would be super lightweight. What do you think @joliss?

PR #251, opened 2015-04-13
skipMissingIndex option to prevent responding with dirTemplate on dir url

bblack:
The use case here is not so much for security (see [here](https://github.com/broccolijs/broccoli/issues/7#issuecomment-32134001)), but to prevent rendering an asset dir when it would conflict with another url. Consider an express app:
```
app.use(broccoliMiddleware);
app.use(sessionMiddleware); // some small middleware that takes non-zero time on each request
app.get('/', function(){ ... });
```
 If we put broccoli first, broccoli will claim the `/` route. If we specify broccoli middleware last, then broccoli will be subject to session middleware, an unnecessary slowdown.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -8,8 +8,26 @@ var mime = require('mime')
 var errorTemplate = handlebars.compile(fs.readFileSync(path.resolve(__dirname, '../templates/error.html')).toString())
 var dirTemplate = handlebars.compile(fs.readFileSync(path.resolve(__dirname, '../templates/dir.html')).toString())
 
+var defaultOptions = {
+  autoIndex: true
+}
+
+function defaults(target, source){
+  // like underscore's _.defaults
+  var out = {}
+  for (key in source) {
+    if (target[key] === undefined) {
+      out[key] = source[key]
+    } else {
+      out[key] = target[key]
+    }
+  }
+  return out
+}
+
 module.exports = function(watcher/* , options */) {
-  var options =  arguments[1] || {};
+  var options = arguments[1] || {};
+  options = defaults(options, defaultOptions)
 
   return function broccoliMiddleware(request, response, next) {
     watcher.then(function(hash) {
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1: from me (I have needed to be able to disable this from Ember CLI when I had an `assets` route in my Ember app).

@joliss - Any objections?

joliss:
Re the implementation:

(1) For the option name, should we make it positive and call it `directoryListings` or `autoindex` (after nginx and Apache) with default to `true`?

(2) When directory listings are disabled, should we also disable slash-redirects, from `/assets` to `/assets/`?

Re the use case:

> If we put broccoli first, broccoli will claim the / route.

Ah yes. I think this is a very reasonable use case - and I'm still torn about @rwjblue's `/assets/` use case. An alternative would be to disable directory listings only for the `/` route. @rwjblue, can we have a chat about this some time to get this resolved?

> If we specify broccoli middleware last, then broccoli will be subject to session middleware, an unnecessary slowdown.

Is this slowdown noticeable? Or are you running Broccoli in production here and want to squeeze out more CPU cycles? I generally advise against running the Broccoli server in production for security reasons but I'd be interested in hearing your use case.

joliss:
> and I'm still torn about @rwjblue's `/assets/` use case.

I'm not opposed to merging this btw, I'd just like to understand the problem/solution space properly.

bblack:
> Is this slowdown noticeable? Or are you running Broccoli in production here and want to squeeze out more CPU cycles? I generally advise against running the Broccoli server in production for security reasons but I'd be interested in hearing your use case.

It's just in development, but it is noticeable--although it's mostly because our middleware is poor. (It runs DB queries on every request and not writing an http response until the queries are done. In development, when requesting many assets, it's probably exhausting the DB connection pool too).

I'll make the option positive as `autoIndex` (like nginx/apache but with conventional js casing), and default it to true.

I have no opinion on the slash-redirect so I won't touch it for now.

joliss:
Landed in a0c9940 (also disables slash-redirects for directories if we're not serving `index.html` or a directory listing), released in 0.15.4 - thanks!

PR #253, opened 2015-04-21
Emit a event to notify that a build is starting.

ShogunPanda:
Hi.

In my fork of broccoli-timepiece (https://github.com/ShogunPanda/broccoli-timepiece) I'm trying to add a file based synchronization mechanism.

My use case is that I run broccoli-timepiece (but this could also apply to `broccoli-cli serve`) in a terminal.
When I go back to Rails I'd like to see a Rails assets pipeline like behavior:

1. The Rails process should not render assets if a build is in progress.
2. The Rails process should show a error template in case of errors.

I already solved the latter by creating a JSON dump in the `error` event and by deleting it in the `change` event.

In this PR I added a `start` event to the Watcher so that I (well, everyone) can be notified when a build starts. The way I'm gonna use it to create a temporary file in the `start` that will be deleted in both the `change` and `error` event.

Let me know if there is a better way to achieve this.
Hope you're gonna like it!
  Paolo

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -10,6 +10,11 @@ function Watcher(builder, options) {
   this.options = options || {}
   this.watchedDirs = {}
 
+  // notify the first build once the object is returned
+  process.nextTick(function(){
+    this.emit('start')
+  }.bind(this))
+
   this.check()
 }
 
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -44,6 +44,7 @@ Watcher.prototype.check = function() {
 
     if (Object.keys(this.watchedDirs).length === 0 || changedDirs.length > 0) {
       this.watchedDirs = {}
+      this.emit('start', changedDirs)
       this.current = this.builder.build(this.addWatchDir.bind(this))
       this.current.then(function(hash) {
         if (this.options.verbose) {


lib/watcher.js
```diff
@@ -10,6 +10,11 @@ function Watcher(builder, options) {
   this.options = options || {}
   this.watchedDirs = {}
 
+  // notify the first build once the object is returned
+  process.nextTick(function(){
+    this.emit('start')
+  }.bind(this))
+
```
joliss: Doesn't the other `emit` in L52 take care of the initial `'start'` event?

joliss:
Seems pretty reasonable to me. If we add it here we should add it to broccoli-sane-watcher as well. @stefanpenner, look good to you?

(Random aside, perhaps we can fold broccoli-sane-watcher back into Broccoli at some point, so we don't have two watchers to maintain.)

stefanpenner:
> (Random aside, perhaps we can fold broccoli-sane-watcher back into Broccoli at some point, so we don't have two watchers to maintain.)

Ya we should likely do this.

stefanpenner:
my advice to the above problem, is to have broccoli serve these assets. No hacky locking required.

lib/watcher.js
ShogunPanda: No.
I don't remember the exact reason (sorry, but I sent this couple of weeks ago XD) but on start up that code is skipped. Probably because no dirs have changed and nothing is being watched. That's why I had to handle that case separately.

ShogunPanda:
Any update on this?

joliss:
Something similar has landed in the meantime:

* In Broccoli 0.16.8, the Builder has `'start'` end `'end`' events ([source](https://github.com/broccolijs/broccoli/blob/e741fece408c4da680b17160971f8638ed853599/lib/builder.js#L44))
* In Broccoli 1.0.0-beta.6, those have been moved to the Watcher, which now has `'build'` and `'change'`/`'error'` events ([source](https://github.com/broccolijs/broccoli/blob/e7eed46100d403fbe1cbb310e992a5b247eb3875/lib/watcher.js#L59))

Thanks again for your PR!

ShogunPanda:
Cool! Thanks! ^^

PR #264, opened 2015-06-04
Add --verbose CLI flag

cecchi:
I added a `--verbose` command line flag to enable verbose output (which currently just prints the slowest trees). It does not seem like the verbose mode should be the default behaviour (as it is now).

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -15,7 +15,7 @@ function broccoliCLI () {
     .description('start a broccoli server')
     .option('--port <port>', 'the port to bind to [4200]', 4200)
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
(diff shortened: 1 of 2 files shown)


joliss:
I want to keep this enabled by default, while most things are in flux and under active development, but eventually we'll want to disable it by default I think. Let's keep this PR open, I'll pull it in at some point.

raiseandfall:
Same thing than #250, closing the one I created a while ago.

stefanpenner:
The code-base has changed quite a-bit.  I think this is a good idea, although I would likely name it `--slow-trees=true|false` and default it to true.

If we wanted completely silent output, we could offer `--silent=true|false` but that is a different and larger effort.

I am closing this for now, but if my above feedback is submitted as a PR, I will help to make sure it lands.

PR #276, opened 2015-08-24
Replace funnel() with new Funnel() in README

alanhogan:
Because that’s the new Broccoli standard. 
https://github.com/broccolijs/broccoli-funnel/commit/038eb2f76423261265e95e0d0a1f87a9aa40accd#commitcomment-12864108

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,9 @@ the following folder within your project folder:
 The following `Brocfile.js` exports the `app/` subdirectory as `appkit/`:
 
 ```js
-var funnel = require('broccoli-funnel')
+var Funnel = require('broccoli-funnel')
 
-module.exports = funnel('app', {
+module.exports = new Funnel('app', {
   destDir: 'appkit'
 })
 ```


alanhogan:
BTW @joliss, I’m also curious about the terminology change from “tree” to “node.” What’s the reasoning there?

rwjblue:
Thank you!

joliss:
> BTW @joliss, I’m also curious about the terminology change from “tree” to “node.” What’s the reasoning there?

Basically I've found that "tree" is a bit confusing to people, so "node" will hopefully make it a bit easier to communicate - people still need to understand that Broccoli describes the build process as a (directed acyclic) graph, but once they do, it should be pretty clear what a "node" is.

alanhogan:
Interesting, @joliss. Previously I’ve thought of “tree” meaning “a directory or tree of files at this point in the build, that can be passed around to other broccoli plugins, composed, and eventually written as the build result.” With “node” described as you have above, my understanding is now of the entire build process being a graph; now, a tree is either the final result or an intermediate step, with previous steps in the build being nodes “beneath” the current node… Hopefully that aligns with your intentions.

PR #279, opened 2015-09-01
hooks to enable outside monitoring

stefanpenner:
lets me do stuff like:

```js
this.builder = new broccoli.Builder(this.tree);

var builder = this;

this.builder.on('start', function(node) {
  builder.monitor = new FSMonitor();
});

this.builder.on('nodeStart', function(node) {
  var metric = new NodeMetric(node);
  builder.monitor.push(metric);
});

this.builder.on('nodeEnd', function(node) {
  builder.monitor.pop();
});

this.builder.on('end', function(node) {
  console.log('endBuild');
  console.log(builder.monitor.totalStats());
});
```

which ultimately results in: https://gist.github.com/stefanpenner/1c319a44a30efd0071fc

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -31,18 +31,24 @@ function summarize(node) {
   }
 }
 
+RSVP.EventTarget.mixin(Builder.prototype);
+
 Builder.prototype.build = function (willReadStringTree) {
   var builder = this
 
   var newTreesRead = []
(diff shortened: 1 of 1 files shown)


joliss:
Released in v0.16.8.

stefanpenner:
yay

PR #283, opened 2015-10-20
`broccoli serve` produces build optionally

chrmod:
`broccoli serve --output=dist`

will create a `dist` folder same as `broccoli build dist` would.

This is required by if you want to "feed in" build into another system during development. 

(in our case it is Firefox extension, so we put build into browser profile and have a debug view served by broccoli at the same time)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /node_modules
 /npm-debug.log
 /test/tmp
+/test/watcher_test.tmp
 /test/multidep
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 logo
 test/tmp
+test/watcher_test.tmp
 test/multidep
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,6 +1,7 @@
 var fs = require('fs')
 var program = require('commander')
 var copyDereferenceSync = require('copy-dereference').sync
+var rimraf = require('rimraf')
 
 var broccoli = require('./index')
 
@@ -26,8 +27,7 @@ function broccoliCLI () {
     .action(function(outputDir) {
       actionPerformed = true
       if (fs.existsSync(outputDir)) {
-        console.error(outputDir + '/ already exists; we cannot build into an existing directory')
-        process.exit(1)
+        rimraf.sync(outputDir)
       }
       var builder = getBuilder()
       builder.build()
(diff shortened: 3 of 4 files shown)


chrmod:
@joliss any comments on this one?

joliss:
I'm not willing to accept 9a669d88e0e0188e3530a1730513db4d918b1c6f (rimraf `outputDir` on `build`), because we'd be rimraf-ing a directory that we didn't create. It's too easy to naively do something like `--output=.` and lose data. Ember CLI uses [heuristics](https://github.com/ember-cli/ember-cli/blob/c6915f683716a6113e8e781adafb0b702244e2e1/lib/models/builder.js#L95-L110) to stop `ember build --output=.` from working, but that is the kind of tricky UI code that I don't want to maintain. Broccoli is a library that's mostly not user facing, so it's not the right place for UI complexity.

As for the main part of this PR (e46e01aaded685bad276d63814d5841dc809d648), I'm inclined to add this feature, but it seems we have three cases:

(1) Continuous serve only (`broccoli serve`)
(2) Continuous serve and build (`broccoli serve --output=dist` in this PR)
(3) Continuous build only ([broccoli-timepiece](https://github.com/rwjblue/broccoli-timepiece))

Can we come up with a coherent CLI syntax that allows for all three?

chrmod:
How about this?

* single build:
`broccoli build dist`

* continuous serve:
`broccoli serve`

* continuous build:
`broccoli watch dist`

* continuous build and serve:
`broccoli serve dist`

It assumes we integrate **broccoli-timepiece** (which name can be considered just a little bit too poetic) into **broccoli-cli**.

chrmod:
fully agreed with rimrafing output folder, this should belong to higher layer.

chrmod:
current implementation that copyDereferenceSync on watcher change is causing as problems as we don't know when copying if finished. Relaying of current cli out can make you restart build dependant process before entire tree is copied which it best case may just not work - in our case of Firefox addon, it causes browser profile to crash.

So copy copy operation should happen before watcher change event is fired, or we should add another event. Which option you would suggest @joliss ?

chrmod:
have moved all project to custom build scripts that use either Server or Warcher JavaScrips APIs, which completely solves my issues and renders this PR obsolete.

chrmod:
the topic got address under new issue https://github.com/broccolijs/broccoli/issues/331

PR #291, opened 2016-01-17
Remove reference to lapsed domain broccoliplugins.com from README

schlueter:
Unless someone plans to renew broccoliplugins.com, it shouldn't be referenced in the README. @pootsbook

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ You can see a full-featured `Brocfile.js` in
 
 ## Plugins
 
-You can find plugins on [broccoliplugins.com](http://broccoliplugins.com) or under the [broccoli-plugin keyword](https://www.npmjs.org/browse/keyword/broccoli-plugin) on npm.
+You can find plugins under the [broccoli-plugin keyword](https://www.npmjs.org/browse/keyword/broccoli-plugin) on npm.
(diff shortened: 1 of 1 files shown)


rwjblue:
Seems good to me. 

@joliss - This ok with you?

pootsbook:
Thanks @schlueter, I had the domain previously, but it looks to have run out behind my back, happy for someone else to pick up the domain or route the broccoli-plugins app to a subdomain on the broccolijs.com domain. The client and server live on Heroku

http://broccoli-plugins-client.herokuapp.com/
http://broccoli-plugins-api.herokuapp.com/

and the code is on GitHub:

https://github.com/emberwatch/broccoli-plugins-search (client)
https://github.com/emberwatch/broccoli-plugins-server (server)

It pretty much runs itself (although you won’t get any data currently because it's hard-coded to broccoliplugins.com) and I would be happy for someone else to take ownership if that was desired, or to shut it down completely if is a better solution. Here are the analytics:

Google Analytics Audience Overview 2015

I can grant full access to anyone with a Google email address.

stefanpenner:
I can take it over: Stefan.penner@gmail.com

PR #292, opened 2016-01-18
Enable enhanced customization of server options

jakedetels:
This pull request aims to enhance the user's ability to customize the options passed to `broccoli.server.serve()` in a couple different ways.

First, users can customize the http response headers created within `lib/middleware.js` by including a `headers` property in the main options object passed in `broccoli.server.serve(builder, options)`.  For example, the following options would enable Broccoli to include a CORS header when serving files on `http://localhost:4200`:

```js
var options = {
  headers: {
    'Access-Control-Allow-Origin': '*'
  }
};

broccoli.server.serve(builder, options);
```

To allow advanced options like these to be easier configure when using Broccoli CLI, this pull request also expands on `lib/cli.js` by allowing users to define an optional `.broccoli` JSON-formatted file within the root directory of their projects.  Any configurations defined within that `.broccoli` file will be pulled whenever calling `broccoli serve`.  For example, the following `.broccoli` file would cause Broccoli to serve on port 12345 with a CORS header:

```js
//  `.broccoli` file saved in project root
{
  "headers": {
    "Access-Control-Allow-Origin": "*"
  },
  "port": 12345
}
```

This pull request includes 5 new passing tests in `test/options_test.js`.  To help clean up the console output when running Mocha, there is new option defined called `disableLogging` which (when set to `true`) will cause `lib/server.js` to *not* call `console.log()` while the Mocha tests are running.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,6 +1,7 @@
 var fs = require('fs')
 var program = require('commander')
 var copyDereferenceSync = require('copy-dereference').sync
+var findup = require('findup-sync')
 
 var broccoli = require('./index')
 
@@ -17,6 +18,7 @@ function broccoliCLI () {
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
     .action(function(options) {
       actionPerformed = true
+      mergeStaticOptions(options);
       broccoli.server.serve(getBuilder(), options)
     })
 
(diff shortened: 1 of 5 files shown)


jakedetels:
@joliss: Any feedback?

oksuz:
@joliss @jakedetels When do you merge this feature? I need this feature for adding custom headers to server response. Thank you

stefanpenner:
@jakedetels / @oksuz sorry for the long delay in response. I believe in broad-strokes I agree with this approach.

The code-base has changed some, if you are still using broccoli and would like to rebase and update the code. I would love to revisit this, and help get it in.

jakedetels:
@stefanpenner: I'll aim to get this PR updated by the end of the month (12/31/2017).  If needed sooner, let me  know.

rwjblue:
Friendly ping 🏂

simonc:
Faced the issue today, this PR would be very helpful 😊

lifeart:
I like idea about additional customization, but I think having multiple ways to do it - not an optimal direction.

We could have `.broccoli.js` (no need to escape comments, able to have dynamic logic in it for different purposes - serve, build-dev, build-prod) file, with all needed things, including middleware, but it's requires more design work here.

PR #296, opened 2016-02-07
add support for serving via ssl

jakedetels:
Allows users to run server using SSL via `broccoli serve -ssl`.

SSL key and certificate files must be located in the user's project root at `ssl/server.key` and `ssl/server.crt`.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -15,6 +15,7 @@ function broccoliCLI () {
     .description('start a broccoli server')
     .option('--port <port>', 'the port to bind to [4200]', 4200)
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
+    .option('--ssl', 'serve via https')
     .action(function(options) {
       actionPerformed = true
       broccoli.server.serve(getBuilder(), options)
--- a/lib/server.js
+++ b/lib/server.js
@@ -21,7 +21,7 @@ function serve (builder, options) {
     try {
       sslOptions = {
         key: fs.readFileSync(path.join(process.cwd(), 'ssl/server.key')),
-        cert: fs.readFileSync(path.join(process.cwd(), 'ssl/server.crt')),
+        cert: fs.readFileSync(path.join(process.cwd(), 'ssl/server.crt'))
       }
     } catch(e) {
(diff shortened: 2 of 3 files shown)


lib/cli.js
```diff
@@ -15,6 +15,7 @@ function broccoliCLI () {
     .description('start a broccoli server')
     .option('--port <port>', 'the port to bind to [4200]', 4200)
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
+    .option('--ssl', 'serve via https')
```
stefanpenner: Lets do:

```
  --ssl (Boolean) (Default: false)
  --ssl-key (String) (Default: ssl/server.key)
  --ssl-cert (String) (Default: ssl/server.crt)
```

lib/server.js
```diff
@@ -9,21 +12,33 @@ function serve (builder, options) {
...
   server.watcher = options.watcher || new Watcher(builder)
 
   server.app = connect().use(middleware(server.watcher))
 
-  server.http = http.createServer(server.app)
+  if (options.ssl) {
+    var sslOptions
+    try {
+      sslOptions = {
+        key: fs.readFileSync(path.join(process.cwd(), 'ssl/server.key')),
```
stefanpenner: `path.join(process.cwd()` here is not needed, `cwd` is implied when using `fs.*` methods

lib/server.js
```diff
@@ -9,21 +12,33 @@ function serve (builder, options) {
...
+        cert: fs.readFileSync(path.join(process.cwd(), 'ssl/server.crt'))
+      }
+    } catch(e) {
+      throw new Error('SSL key and certificate files must be located at ssl/server.key and ssl/server.crt')
```
stefanpenner: this error message likely should include `cwd` in the errors.

stefanpenner:
@jakedetels sorry for the late review, if you can make the above changes. We can merge and :shipit:

jakedetels:
@stefanpenner: I'll update this by end-of-month..

rwjblue:
Friendly ping 😺

stefanpenner:
landed via https://github.com/broccolijs/broccoli/pull/417

PR #306, opened 2016-06-25
cli: serve optionally creates build

chrmod:
follow up for https://github.com/broccolijs/broccoli/pull/283#issuecomment-228538964


chrmod:
Here you can find a backport for 0.16.x https://github.com/cliqz-oss/broccoli/tree/serve-build-16

PR #312, opened 2016-11-21
POC [Fixes #311] fix cleanup

### LINKED ISSUE #311: mid-build <ctrl-c> doesn't always cleanup directories.
Will confirm in isolation, but it seems like the relevant code *may* not support this (yet).

stefanpenner:
* ensure cleanup gives the ability to wait until it is done
* improves support for mid-build cleanup, waits for the current task to complete, prevents the next task, and then exists.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -53,6 +53,7 @@ function Builder(outputNode, options) {
 
   this.setupTmpDirs()
 
+  this._currentStep = RSVP.resolve()
   // Now that temporary directories are set up, we need to run the rest of the
   // constructor in a try/catch block to clean them up if necessary.
   try {
@@ -76,9 +77,12 @@ RSVP.EventTarget.mixin(Builder.prototype)
 // This method will never throw, and it will never be rejected with anything
 // other than a BuildError.
 Builder.prototype.build = function() {
+  if (this._canceled) {
+    throw new Error('This Builder cannot be built, as it has been canceled');
+  }
   var self = this
   this.buildId++
-  var promise = RSVP.resolve()
+  var promise = RSVP.resolve();
   this.nodeWrappers.forEach(function(nw) {
     // We use `.forEach` instead of `for` to close nested functions over `nw`
 
@@ -87,6 +91,7 @@ Builder.prototype.build = function() {
 
     promise = promise
       .then(function() {
+        if (self.cancelled) { return }
         // We use a nested .then/.catch so that the .catch can only catch errors
         // from this node, but not from previous nodes.
         return RSVP.resolve()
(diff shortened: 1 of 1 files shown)


lib/builder.js
```diff
@@ -104,13 +111,15 @@ Builder.prototype.build = function() {
...
   })
-  return promise
+  return promise;
 }
 
 // Destructor-like method. Cleanup is synchronous at the moment, but in the
 // future we might change it to return a promise.
 Builder.prototype.cleanup = function() {
+  this._canceled = true
   this.builderTmpDirCleanup()
```
joliss: I don't think it's correct to try removing tmp dirs with `builderTmpDirCleanup` while the build is still in progress. We'd need to wait for the current step to finish first.

lib/builder.js
```diff
@@ -87,14 +91,17 @@ Builder.prototype.build = function() {
 
     promise = promise
       .then(function() {
+        if (self.cancelled) { return }
```
joliss: Tests will catch this anyway, but there's a tyop here (`self._canceled`) :)

lib/builder.js
joliss: Oh. Wait, I think you're already saying this in your original comment (your list is a to-do list, yes?)...

joliss:
I actually wonder if it's conceptually necessary to tie the cancellation thing to the `cleanup` function. In other words, I don't think there's any reason why in general we couldn't start a fresh build after canceling the current build, instead of cleaning up immediately. We shouldn't end up in any "wedged" state that would stop us from doing so.

So if we do that, the Ctrl+C handler would do something like:

```js
cancelCurrentBuild()
then wait for currentBuild
then cleanup
```

lib/builder.js
stefanpenner: ya, the version in broccoli-builder 0.18.x waits

stefanpenner:
@joliss i'll update this to be more like what I ended-up with on 0.18.x. There are some nuanced differences, that have not made it this far yet.

---

* adding a public `cancel` independent of cleanup is totally fine, will gladly add that
* cleanup should likely still call cancel, otherwise cleanup and an running build will simply race
* we need to decide the ordering between the completion of `cancel()` `cleanup()` and `builder.build()`. I believe `builder.build` completion must wait for any pending `cleanup()` which in tern must wait for any pending `cancel()`. That way the external world remains observably consistent, regardless if a build completes via rejection due to cancellation, due to error, or totally normal runtime.

stefanpenner:
replaced by: https://github.com/broccolijs/broccoli/pull/350

PR #318, opened 2017-02-02
Add typescript types

krisselden:


joliss:
I don't want to maintain TypeScript definitions in parallel. So I wouldn't want to merge this unless I can actually use TypeScript as part of my development workflow for Broccoli (I haven't tried).

krisselden:
@joliss I'm happy to maintain it and/or would you be open to using ts for broccoli?

krisselden:
I'm happy to convert things, it would be helpful to have well defined interfaces.  I'm also happy to keep declarations in sync.

rwjblue:
@krisselden - We should look at this again. If we want to publish types, we should just rewrite in typescript....

oligriffiths:
@krisselden see https://github.com/broccolijs/broccoli/pull/390 ;)

stefanpenner:
the project has been rewritten in typescript, i suspect we can close this now.

PR #319, opened 2017-06-08
update dependencies

bekzod:


chrmod:
@stefanpenner @joliss this PR broke master. Update of `broccoli-slow-trees` relies on `heimdalljs` to provide timing information. This results in this error:
```
Error when printing slow nodes: TypeError: tree.visitPostOrder is not a function
``` 
once trying to `broccoli serve` any build tree.

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/broccolijs/broccoli",
   "dependencies": {
     "broccoli-node-info": "1.1.0",
-    "broccoli-slow-trees": "^2.0.0",
+    "broccoli-slow-trees": "3.0.1",
```
Turbo87: @bekzod why did you pin this dependency instead of using a caret constraint?

bekzod: because I am crazy :D .... tbh I think it was accident

bekzod: sorry

PR #327, opened 2017-11-22
Fix server_test’s hanging without mocha —exit

stefanpenner:


rwjblue:
I believe the test failure is due to https://github.com/broccolijs/broccoli/blob/d6ca81689471eb6212a76078d2ac92a6cc30c09b/lib/server.js#L67

PR #328, opened 2017-11-22
Test Watcher

stefanpenner:
* fix some potential race conditions
* fix some leaks (this.readyPromises etc)
* don’t leak internals. For example watch yielded an array of watchers, likely by mistake.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "findup-sync": "^2.0.0",
     "handlebars": "^4.0.11",
     "heimdalljs-logger": "^0.1.9",
-    "mime": "^1.3.4",
+    "mime": "^1.5.0",
     "rimraf": "^2.6.2",
     "rsvp": "^4.7.0",
     "sane": "^2.2.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "findup-sync": "^2.0.0",
     "handlebars": "^4.0.11",
     "heimdalljs-logger": "^0.1.9",
-    "mime": "^1.3.4",
+    "mime": "1.5.0",
```
rwjblue: Seems odd to hard code to this specific version, did you mean `^1.5.0`?

rwjblue:
Windows failure is:

```
1) WatcherAdapter
       watch
         actually works !!:
     Error: EINVAL: invalid argument, utime 'C:\projects\broccoli\test\fixtures\basic\foo.txt'
      at Error (native)
      at Object.fs.utimesSync (fs.js:1077:11)
      at test\watch_adapter_test.js:134:12

```

rwjblue:
Random snippet [from the internet](https://github.com/nodejs/node/issues/5561#issuecomment-192306799) which suggests a fix for this:

> atime and mtime parameters must be instances of Date().

Basically:

```js
// Get current date
var currentDate = new Date();

var modifiedDate = new Date(modifiedUnixTimestamp * 1000);

var fullPathOfFile = "C:/path/to/file.txt";

// File, created date, modified date
fs.utimesSync(fullPathOfFile, currentDate, modifiedDate);
```

stefanpenner:
Thanks for looking. I actually have that fix locally, but apparently forgot to push. Will push when I’m near my computer.

Silly windows

Issue #331, opened 2017-11-23
CLI improvements


stefanpenner:
## Guiding Ideas:

broccoli's cli should have:

* a simple implementation (complex features should be avoided)
* a simple developer experience (not all the features, but the ones that matter)

### Stuff
- [x] `broccoli help` and `broccoli help serve` should display various serve options --port --host (and in the future more
- [x] build and serve should have more options, for example:
  - [x] `--brocfile-path=` for running specific brocfiles (@chrmod)
  - [x] `--cwd=` for specifying the cwd of a given brocfile
  - [x] `--output-path=`
  - [x] `--watcher=` allowing configuration of which watch type should be used. Some users may need polling due to FS limitations etc. options = polling, watchman, events
  - [x] `--watch=` allow toggling the watcher (for both build and watch) #

chrmod:
@stefanpenner what `--watch` option should do for `build` command?

stefanpenner:
> @stefanpenner what --watch option should do for build command?

`build --watch` should be the same as `serve` but without HTTP server.

In theory we may just want to have a `build` that optionally:
* can serve
* can build
* can rebuild

chrmod:
@stefanpenner any suggestion how `help` should change its behavior? do we look for more informative descriptions or rather stylistic improvements like colors?

stefanpenner:
@chrmod I believe help should enable discovery of functionality. For example `help serve` should inform us of the various options/arguments.

chrmod:
@stefanpenner 

those are the outputs that `commander.js` currently generates:
* `broccoli`
```

  Usage: broccoli [options] <command> [<args ...>]


  Options:

    -V, --version  output the version number
    -h, --help     output usage information


  Commands:

    serve|s [options]         start a broccoli server
    build [options] [target]  output files to target directory
```
* `broccoli build --help`
```

  Usage: build [options] [target]

  output files to target directory


  Options:

    --brocfile-path <path>  the path to brocfile
    --output-path <path>    the path to target output folder
    --watch                 turn on the watcher
    --watcher <watcher>     select sane watcher mode
    -h, --help              output usage information
```
* `broccoli serve --help`
```
  Usage: serve|s [options]

  start a broccoli server


  Options:

    --port <port>           the port to bind to [4200] (default: 4200)
    --host <host>           the host to bind to [localhost] (default: localhost)
    --brocfile-path <path>  the path to brocfile
    --output-path <path>    the path to target output folder
[...]

PR #337, opened 2017-11-26
cli: brocfile path option

chrmod:
implements part of #331

### DIFF
--- a/lib/load_brocfile.js
+++ b/lib/load_brocfile.js
@@ -4,10 +4,16 @@ const path = require('path');
 const findup = require('findup-sync');
 
 module.exports = loadBrocfile;
-function loadBrocfile() {
-  const brocfile = findup('Brocfile.js', {
-    nocase: true,
-  });
+function loadBrocfile(brocfilePath) {
+  let brocfile;
+
+  if (brocfilePath) {
+    brocfile = path.resolve(brocfilePath);
+  } else {
+    brocfile = findup('Brocfile.js', {
+      nocase: true,
+    });
+  }
 
   if (brocfile == null) throw new Error('Brocfile.js not found');
 
--- a/test/load_brocfile_test.js
+++ b/test/load_brocfile_test.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const loadBrocfile = require('../lib/load_brocfile');
+const chai = require('chai');
+
+const projectPath = 'test/fixtures/project';
+const brocfileFixture = require('../' + projectPath + '/Brocfile.js');
+
+describe('loadBrocfile', function() {
+  let oldCwd = null;
+
+  beforeEach(function() {
+    oldCwd = process.cwd();
+  });
+
+  afterEach(function() {
+    process.chdir(oldCwd);
+  });
+
+  context('without Brocfile.js', function() {
+    beforeEach(function() {
+      process.chdir('test/fixtures/basic');
+    });
+
+    it('return tree definition', function() {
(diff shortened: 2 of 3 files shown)


chrmod:
@stefanpenner I'm happy to implement the rest of #331

lib/load_brocfile.js
```diff
@@ -4,10 +4,16 @@ const path = require('path');
 const findup = require('findup-sync');
 
 module.exports = loadBrocfile;
-function loadBrocfile() {
-  const brocfile = findup('Brocfile.js', {
-    nocase: true,
-  });
+function loadBrocfile(brocfilePath) {
```
stefanpenner: lets add some unit tests to this function to ensure we don't regress

stefanpenner:
> @stefanpenner I'm happy to implement the rest of #331

I would love that!

test/load_brocfile_test.js
```diff
@@ -0,0 +1,49 @@
...
+      chai.expect(loadBrocfile()).to.equal(brocfileFixture);
+    });
+  });
+
+  context('with Brocfile.js, called in subfolder', function() {
+    beforeEach(function() {
+      process.chdir(projectPath + '/subdir');
+    });
+
+    afterEach(function() {
```
stefanpenner: i dont think this `afterEach` is required, as the parent describe blocks afterEach already restores the `chdir`

chrmod:
@stefanpenner alright. Guess more tests for `load_brocfile` will arrive with `cwd` option.

test/load_brocfile_test.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    it('return tree definition', function() {
+      chai.expect(loadBrocfile()).to.equal(brocfileFixture);
+    });
+  });
+
+  context('with path', function() {
+    it('return tree definition', function() {
+      chai.expect(loadBrocfile(projectPath + '/Brocfile.js')).to.equal(brocfileFixture);
```
stefanpenner: can we also test the error case here? Specifically that if an explicit path is given, but that path doesn't contain a file that we hit the helpful error?

test/load_brocfile_test.js
chrmod: it is needed as `load_brocfile` changes working dir https://github.com/broccolijs/broccoli/blob/master/lib/load_brocfile.js#L18

test/load_brocfile_test.js
chrmod: make sense :+1:

test/load_brocfile_test.js
stefanpenner: but doesn't the outer `afterEach` already `chdir`? Am I confused and that one doesn't fire?

test/load_brocfile_test.js
stefanpenner: thx

test/load_brocfile_test.js
chrmod: right, sorry, it's pretty late here

test/load_brocfile_test.js
chrmod: done

test/load_brocfile_test.js
stefanpenner: no worries

stefanpenner:
LGTM

stefanpenner:
rebasing

stefanpenner:
rebased -> https://github.com/broccolijs/broccoli/pull/339

PR #338, opened 2017-11-26
quick fix for cli build command

chrmod:
current master throws:
`TypeError: builder.cleanup is not a function`

### DIFF
--- a/test/cli_test.js
+++ b/test/cli_test.js
@@ -1,10 +1,12 @@
 'use strict';
 
+const Builder = require('../lib/builder');
 const broccoli = require('../lib/index');
 const chai = require('chai');
+const cli = require('../lib/cli');
+const rimraf = require('rimraf');
 const sinon = require('sinon').createSandbox();
 const sinonChai = require('sinon-chai');
-const cli = require('../lib/cli');
 
 chai.use(sinonChai);
 
@@ -19,6 +21,40 @@ describe('cli', function() {
   afterEach(function() {
     sinon.restore();
     process.chdir(oldCwd);
+  });
+
+  describe('build', function() {
+    let exitStub;
+
+    beforeEach(function() {
+      exitStub = sinon.stub(process, 'exit');
+    });
+
+    afterEach(function() {
+      rimraf.sync('dist');
+    });
+
+    context('on successful build', function() {
+      it('cleanups tmp files', function(done) {
+        const cleanup = sinon.spy(Builder.prototype, 'cleanup');
+
+        cli(['node', 'broccoli', 'build', 'dist']);
+
+        process.nextTick(() => {
+          chai.expect(cleanup).to.be.calledOnce;
+          done();
+        });
+      });
+
+      it('closes process on completion', function(done) {
+        cli(['node', 'broccoli', 'build', 'dist']);
+
+        process.nextTick(() => {
+          chai.expect(exitStub).to.be.calledWith(0);
+          done();
+        });
+      });
+    });
   });
 
   describe('serve', function() {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /node_modules
 /npm-debug.log
 /test/multidep
+/test/fixtures/project/dist
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -40,7 +40,9 @@ module.exports = function broccoliCLI(args) {
         process.exit(1);
       }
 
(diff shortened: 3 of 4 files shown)


chrmod:
@rwjblue make sense. I'm on it

stefanpenner:
Apparently, I don't understand the TDZ.

chrmod:
@rwjblue @stefanpenner those tests were not too easy to write and I'm not 100% sure if those are what we want from them. Please review.

test/cli_test.js
```diff
@@ -21,6 +23,40 @@ describe('cli', function() {
...
+    });
+
+    context('on successful build', function() {
+      it('cleanups tmp files', function(done) {
+        const cleanup = sinon.spy(Builder.prototype, 'cleanup');
+
+        cli(['node', 'broccoli', 'build', 'dist']);
+
+        process.nextTick(() => {
+          chai.expect(cleanup).to.be.calledOnce;
```
stefanpenner: If this assertion fails, i believe this test wont fail with a useful error message, rather it will time out.
Remember in mocha, assertions are thrown exceptions. That means the `done` will never be hit...

We will need to address that before we can merge this.

One option would be:

```js
process.nextTick(() => {
  try {
    chai.expect(cleanup).to.be.calledOnce;
    done();
  } catch(e) {
    done(e); // fail the test with the given error
  }
});
```

test/cli_test.js
```diff
@@ -21,6 +23,40 @@ describe('cli', function() {
...
+          done();
+        });
+      });
+
+      it('closes process on completion', function(done) {
+        cli(['node', 'broccoli', 'build', 'dist']);
+
+        process.nextTick(() => {
+          chai.expect(exitStub).to.be.calledWith(0);
```
stefanpenner: If this assertion fails, i believe this test wont fail with a useful error message, rather it will time out.
Remember in mocha, assertions are thrown exceptions. That means the `done` will never be hit...

We will need to address that before we can merge this.

One option would be:

```js
process.nextTick(() => {
  try {
    chai.expect(cleanup).to.be.calledOnce;
    done();
  } catch(e) {
    done(e); // fail the test with the given error
  }
});
```

test/cli_test.js
stefanpenner: > process.nextTick(() => {

It's a bummer we need to do this, but commander doesn't make this stuff very easy to test.

chrmod: it's super interesting to me now as I you're 100% right with try/catch solution. so I have red/green tested this code by replacing `calledOnce` with `calledTwice` and it gave me perfectly fine error message
```
  1) cli build on successful build cleanups tmp files:                                                                                                                                        
     Uncaught expected cleanup to have been called exactly twice, but it was called once                                                                                                      
    cleanup() at builder.build.then.finally (/home/chrmod/Projects/chrmod/broccoli/lib/cli.js:48:32)   
```
so I've remove `done()` call, and it still worked.

Does mocha somehow handles `process.nextTick`? (`done` argument still have to be present for it to work).

test/cli_test.js
chrmod: maybe this behavior is somehow related to node version. If `nextTick` would be handled synchronously this would not have to be an issue. 

I'm open to suggestions how to handle those tests. 

Easy solution is to refactor `lib/cli` module and cover it with unit tests. But there is quite some value in full integration tests for command line tools, so I'm not sure.

test/cli_test.js
chrmod: there is a mocha test case that does the same https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/test/integration/fixtures/uncaught-hook.fixture.js#L10

guess it is fine to keep it like that

chrmod:
@stefanpenner can you please replay appveyor. Wonder if it is random fail or something is different on node 9.

test/cli_test.js
stefanpenner: interesting, guess they made this work.

stefanpenner:
> @stefanpenner can you please replay appveyor. Wonder if it is random fail or something is different on node 9.

Unfortunately I cannot, appveyor does not believe I am an owner of this repo.... for unknown reasons.

rwjblue:
Same here. 🤔

stefanpenner:
I'll merge, and keep an eye on master.

@chrmod much appreciated!!

PR #340, opened 2017-11-28
Fixing cli bugs for --brocfile param

chrmod:
So my previous PR was not so good in the end. Here goes fixes and some tests to cover them.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -27,7 +27,7 @@ module.exports = function broccoliCLI(args) {
     .action(options => {
       actionPerformed = true;
 
-      const builder = getBuilder(options.brocfilePath);
+      const builder = getBuilder({ brocfilePath: options.brocfilePath });
 
       broccoli.server.serve(new Watcher(builder), options.host, parseInt(options.port, 10));
     });
--- a/test/cli_test.js
+++ b/test/cli_test.js
@@ -4,6 +4,7 @@ const Builder = require('../lib/builder');
 const broccoli = require('../lib/index');
 const chai = require('chai');
 const cli = require('../lib/cli');
+const loadBrocfile = require('../lib/load_brocfile');
 const rimraf = require('rimraf');
 const sinon = require('sinon').createSandbox();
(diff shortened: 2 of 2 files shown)


lib/cli.js
```diff
@@ -23,19 +23,19 @@ module.exports = function broccoliCLI(args) {
     .description('start a broccoli server')
     .option('--port <port>', 'the port to bind to [4200]', 4200)
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
-    .option('--brocfile-path <path>', 'the path to brocfile')
+    .option('--brocfile <path>', 'the path to brocfile')
```
rwjblue: We decided to change this to `--brocfile-path`, can you bring that back?

chrmod: done. can you please update https://github.com/broccolijs/broccoli/issues/331 to reflect your decisions. 
I plan to implement other params now, so would like to be sync :-)

rwjblue: No problem!  I updated it (we only discussed `--brocfile-path` though)...

lib/cli.js
```diff
@@ -27,7 +27,7 @@ module.exports = function broccoliCLI(args) {
     .action(options => {
       actionPerformed = true;
 
-      const builder = getBuilder(options.brocfilePath);
+      const builder = getBuilder({ brocfilePath: options.brocfilePath });
```
rwjblue: I'm unsure why we need to make a second object just to pass the single option into it. Any reason not to do either `getBuilder(options)` or `getBuilder(options.brocfilePath)` here?

chrmod: was planning to pass more options here, like `cwd`. But have to see how other options will be passed around to builder/watcher. As `build` and `serve` are going to share options, probably this is going to change.

chrmod: if you prefer I can simplify it now and change in upcoming PRs

rwjblue: Nah, seems totally fine, I was mostly just curious.  Answer is: planning more changes that will leverage the object. 👍

PR #341, opened 2017-11-28
Cli option --output-path

chrmod:
Implements another part of https://github.com/broccolijs/broccoli/issues/331

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const TreeSync = require('tree-sync');
-const copyDereferenceSync = require('copy-dereference').sync;
 const fs = require('fs');
 
 const broccoli = require('./index');
@@ -34,10 +33,7 @@ module.exports = function broccoliCLI(args) {
       const watcher = new Watcher(builder);
 
       if (outputDir) {
-        if (fs.existsSync(outputDir)) {
-          console.error(outputDir + '/ already exists; we cannot build into an existing directory');
-          process.exit(1);
-        }
+        guardOutputDir(outputDir);
         const outputTree = new TreeSync(builder.outputPath, outputDir);
 
         watcher.on('buildSuccess', function() {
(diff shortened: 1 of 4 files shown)


lib/cli.js
```diff
@@ -24,21 +25,46 @@ module.exports = function broccoliCLI(args) {
...
+      const watcher = new Watcher(builder);
+
+      if (outputDir) {
+        if (fs.existsSync(outputDir)) {
+          console.error(outputDir + '/ already exists; we cannot build into an existing directory');
+          process.exit(1);
+        }
+
+        watcher.on('buildSuccess', function() {
+          rimraf.sync(outputDir);
```
stefanpenner: often people symlink to this directory, deleting it will break that.

in ember-cli we use: https://www.npmjs.com/package/tree-sync for relatively efficient syncing. Thoughts?

chrmod: have some projects that also symlink to output directory and they work fine with serve, one example of Firefox extension: https://github.com/cliqz-oss/browser-core/blob/master/fern/commands/serve.js#L59

Some watches or build system can get confused, I've seen it with `gulp`, but `webpack` in contrary is fine with symlink that get deleted and recreated. 

Did not work with `tree-sync` yet, but looks as optimal solution to the problem (no need for full copy if only some files were changed). 

Should I try to introduce it in this PR?

stefanpenner: > Should I try to introduce it in this PR?

Ya. We needed to add this to ember-cli, as some users had very large output dirs, (videos/pictures and stuff) and switching to tree-sync fix some serious rebuild issues.

chrmod: done. going to use `tree-sync` in my other projects, it works great.

lib/cli.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
-const fs = require('fs');
+const TreeSync = require('tree-sync');
 const copyDereferenceSync = require('copy-dereference').sync;
```
rwjblue: Can you swap the remaining usage of `copy-dereference` to use `tree-sync` (I believe it is in `build` command below)? Then we can remove it as a dependency...

lib/cli.js
```diff
@@ -24,21 +25,46 @@ module.exports = function broccoliCLI(args) {
...
+    .option('--output-path <path>', 'the path to target output folder')
     .action(options => {
       actionPerformed = true;
 
+      const outputDir = options.outputPath;
       const builder = getBuilder({ brocfilePath: options.brocfilePath });
+      const watcher = new Watcher(builder);
+
+      if (outputDir) {
+        if (fs.existsSync(outputDir)) {
```
rwjblue: Can we make a shared function in module scope for this assertion? It seems to be the same between here and the `build` command...

chrmod: done

chrmod:
@rwjblue @stefanpenner can we please push it forward? 
also https://github.com/broccolijs/broccoli/pull/342

Issue #184, opened 2014-08-12
`broccoli watch` could still be useful


jlongster:
I couldn't find a previous issue about this, so asking here. You say "no more `grunt watch`", but really to me broccoli just gets the dependency handling correct, and enables a much better "watch". Right now the broccoli-cli only supports `build` and `serve` commands; what harm would a `watch` command have, which is exactly like `serve` except instead of firing up a webserver, it writes the final file structure to dist on every change?

fredrikekelund:
I agree completely that `broccoli watch` would be very useful! `broccoli serve` can complicate interaction with back-ends via relative URL's, so watch functionality together with livereload would be :ok_hand:

There is [broccoli-timepiece](https://github.com/rwjblue/broccoli-timepiece) which does provide a watch like command. It lacks livereload functionality though, but I'm trying to push for that here https://github.com/rwjblue/broccoli-timepiece/issues/9 And in any case, I do think the watch command should be a part of broccoli core.

rwjblue:
@fredrikekelund - I disagree, `watch` should absolutely be facilitated by the core library but not bundled with it.  I believe that we should remove nearly all of the CLI code from this repo, and move that into broccoli-cli package (which is already required).

fredrikekelund:
Sorry, my bad - that should be bundled with broccoli-cli, not necessarily with broccoli core. Basically what I was looking for was that it would make sense to be able to run `broccoli watch dest`. What do you think @rwjblue?

stefanpenner:
I'n ember-cli we added `ember build --watch` and it has proved more useful then expected. We should likely consider added it here.

fredrikekelund:
@stefanpenner with Watcher already being bundled in core and error reporting already in place for the build CLI command, I think this would be trivial to implement.

Would you accept a PR if I took a stab at it?

chrmod:
@fredrikekelund I'm also on same issue, implementing https://github.com/broccolijs/broccoli/issues/331 step by step. 
If you're fine with it, I should be able to finish it before end of the week.

fredrikekelund:
@chrmod ah, that's great! No worries, just wanted to offer my assistance

PR #342, opened 2017-11-29
Cli option --watch

chrmod:
To be merged after: https://github.com/broccolijs/broccoli/pull/341
Implements another part of: https://github.com/broccolijs/broccoli/issues/331
Closes: https://github.com/broccolijs/broccoli/issues/184

Notable changes:
* build/serve share some console output encapsulated in `lib/messages.js`
* `lib/cli.js` always return the promise to simplify testing (no more `nextTick`)

Questions: 
* listening to OS signals is duplicated in `lib/cli.js` and `lib/server.js` - I think we can remove it from server completely as it should probably be hosting program responsibility (in this case the cli)
* `lib/dummy-watcher.js` is a way to make `lib/server.js` run correctly without code reloading, but the name is not very descriptive and it is not clear if it should be a part of public interface

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -4,7 +4,6 @@ const TreeSync = require('tree-sync');
 const fs = require('fs');
 
 const broccoli = require('./index');
-const Watcher = require('./watcher');
 
 module.exports = function broccoliCLI(args) {
   // always require a fresh commander, as it keeps state at module scope
@@ -25,8 +24,11 @@ module.exports = function broccoliCLI(args) {
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
     .option('--brocfile-path <path>', 'the path to brocfile')
     .option('--output-path <path>', 'the path to target output folder')
+    .option('--no-watch', 'turn off the watcher')
     .action(options => {
       actionPerformed = true;
+
+      const Watcher = options.watch ? broccoli.Watcher : broccoli.DummyWatcher;
 
       const outputDir = options.outputPath;
       const builder = getBuilder({ brocfilePath: options.brocfilePath });
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -27,7 +27,7 @@ module.exports = function broccoliCLI(args) {
     .option('--output-path <path>', 'the path to target output folder')
     .option('--no-watch', 'turn off the watcher')
     .action(options => {
-      const Watcher = options.watch ? broccoli.Watcher : broccoli.DummyWatcher;
+      const Watcher = getWatcher(options.watch);
       const outputDir = options.outputPath;
       const builder = getBuilder({ brocfilePath: options.brocfilePath });
       const watcher = new Watcher(builder);
@@ -52,7 +52,7 @@ module.exports = function broccoliCLI(args) {
     .option('--output-path <path>', 'the path to target output folder')
     .option('--watch', 'turn on the watcher')
     .action((outputDir, options) => {
-      const Watcher = options.watch ? broccoli.Watcher : broccoli.DummyWatcher;
+      const Watcher = getWatcher(options.watch);
 
       if (outputDir && options.outputPath) {
         console.error('option --output-path and [target] cannot be passed at same time');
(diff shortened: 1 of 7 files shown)


chrmod:
@stefanpenner @rwjblue please review as some solutions here may be questionable.

lib/cli.js
```diff
@@ -24,50 +24,96 @@ module.exports = function broccoliCLI(args) {
     .option('--port <port>', 'the port to bind to [4200]', 4200)
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
     .option('--brocfile-path <path>', 'the path to brocfile')
+    .option('--output-path <path>', 'the path to target output folder')
+    .option('--no-watch', 'turn off the watcher')
```
rwjblue: I think having `--watch` as a boolean arg would by default also support `--no-watch`.

Also, I think we should add a `--watcher` option which takes a string to specify the type of watcher to use.  In ember-cli we support `polling`, `node`, and `watchman` which is essentially just passed through to `sane` ([see sane README for details](https://github.com/amasad/sane)).

lib/cli.js
```diff
@@ -24,50 +24,96 @@ module.exports = function broccoliCLI(args) {
...
+      actionPromise = (server && server.closingPromise) || RSVP.resolve();
     });
 
   program
-    .command('build <target>')
+    .command('build [target]')
     .description('output files to target directory')
     .option('--brocfile-path <path>', 'the path to brocfile')
+    .option('--output-path <path>', 'the path to target output folder')
+    .option('--watch', 'turn on the watcher')
```
rwjblue: Same as above, can we also add a `--watcher` option?

lib/index.js
```diff
@@ -13,6 +13,9 @@ module.exports = {
   get getMiddleware() {
     return require('./middleware');
   },
+  get DummyWatcher() {
```
rwjblue: Not sure we want a `DummyWatcher` exposed here, it seems like its only used/needed internally when `--no-watch` is specified. In that case, we _probably_ don't need to export it here...

lib/cli.js
chrmod: going to implement `--watcher` in separate PR

lib/index.js
chrmod: okay

lib/cli.js
chrmod: `--watch` defaults to `options.watch === undefined`
while
`--no-watch` defaults to `options.watch === true`

so I think it make sense to keep `--no-watch` on `serve` and `--watch` on `build.

Please take a look on commit https://github.com/broccolijs/broccoli/pull/342/commits/0930a3c5d4dbc6d8ad58599368a6217d156795a4

chrmod:
Some of the tests does not close the server and hang. Need to detect which one it is but as it happens only in CI I will be pushing commits to this PR to detect problematic test. Please do not merge even if PR is green.

chrmod:
okay, wanted to cover the case on which server is closed on user input (SIGTERM), but the test is not working correctly in CI. Thus I remove it from the suite. Maybe someone can suggest how to approach the problem.

Problematic test:
```js
    context('on terminated watcher', function() {
      it('by SIGTERM exits without error', function(done) {
        const closingPromise = cli(['node', 'broccoli', 'serve']);
        setTimeout(() => {
          process.kill(process.pid, 'SIGTERM');
        }, 200);
        process.once('SIGTERM', () => {
          closingPromise
            .then(() => {
              chai.expect(exitStub).to.be.calledWith(0);
              done();
            })
            .catch(done);
        });
      });
    });
```

stefanpenner:
LGTM, sorry for the delay was AFK for a-few days.

@chrmod this looks good to ship, can you confirm? (I'll merge once you confirm)

chrmod:
@stefanpenner it is fine to merge, thanks!

danielo515:
How can I make use of this ? There is no reference on the cli help and any option I tried ends on a unknown option `--watch`

PR #344, opened 2017-12-04
Cli option cwd

chrmod:
to be merged after: https://github.com/broccolijs/broccoli/pull/342

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -25,11 +25,12 @@ module.exports = function broccoliCLI(args) {
     .option('--host <host>', 'the host to bind to [localhost]', 'localhost')
     .option('--brocfile-path <path>', 'the path to brocfile')
     .option('--output-path <path>', 'the path to target output folder')
+    .option('--cwd <path>', 'the path to working folder')
     .option('--no-watch', 'turn off the watcher')
     .option('--watcher <watcher>', 'select sane watcher mode')
     .action(options => {
-      const builder = getBuilder({ brocfilePath: options.brocfilePath });
-      const Watcher = getWatcher(options.watch);
+      const builder = getBuilder(options);
+      const Watcher = getWatcher(options);
       const outputDir = options.outputPath;
       const watcher = new Watcher(builder, buildWatcherOptions(options));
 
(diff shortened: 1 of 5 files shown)


lib/cli.js
```diff
@@ -63,38 +67,62 @@ module.exports = function broccoliCLI(args) {
...
+  return actionPromise || RSVP.resolve();
 };
 
 function getBuilder(options) {
   if (!options) {
     options = {};
   }
-  return new broccoli.Builder(broccoli.loadBrocfile(options.brocfilePath));
+  const brocfile = broccoli.loadBrocfile(options.brocfilePath);
+  process.chdir(options.cwd || brocfile.baseDir);
```
stefanpenner: I would have thought we would `chdir` before loading the brocfile, that way `cwd` is consistent before we touch user code.

chrmod: @stefanpenner was thinking about that and have mixed feelings. 

consider using `--brocfile-path` with `--cwd` together, example:
`broccoli serve --cwd <project-path> --brocfile-path ./Brocfile.development.js`
should the brocfile path be resolved to `<project-path>/Brocfile.development.js` or to `<current working dir>/Brocfile.development.js` ? 


another issue is `findup` in `load_brocfile.js` - will it be understandable that findup will be affected by `--cwd`?

stefanpenner: hmm, I'll have to think a bit more..

stefanpenner: I think `--cwd` should not affect the CLI arguments, but should affect things like `process.cwd()` in userland code. My mental model is that CLI parameters are evaluated in the context of the shell. This ensures the shells autocomplete works regardless of the `--cwd`. As if the user wants to change `cwd` in the shell, they should just do that...

So:

```sh
broccoli --cwd somewhere/else --brocfile-path ./Brocfile.development.js
```

```js
// Brocfile.js
process.cwd() // should be --cwd
```

chrmod: :+1: I do like the argument with shell autocompletion

chrmod:
@stefanpenner I've updated the PR to change `cwd` behavior. Still have to update tests to make some sense. Please suggest if we should test something specific.

chrmod:
@stefanpenner actually those tests kinda work. Please consider a merge

rwjblue:
Thanks @chrmod !

PR #345, opened 2017-12-07
cli: option --watcher

chrmod:
Implements another part of https://github.com/broccolijs/broccoli/issues/331

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -26,11 +26,12 @@ module.exports = function broccoliCLI(args) {
     .option('--brocfile-path <path>', 'the path to brocfile')
     .option('--output-path <path>', 'the path to target output folder')
     .option('--no-watch', 'turn off the watcher')
+    .option('--watcher <watcher>', 'select sane watcher mode')
     .action(options => {
       const Watcher = getWatcher(options.watch);
       const outputDir = options.outputPath;
       const builder = getBuilder({ brocfilePath: options.brocfilePath });
-      const watcher = new Watcher(builder);
+      const watcher = new Watcher(builder, buildWatcherOptions(options));
 
       if (outputDir) {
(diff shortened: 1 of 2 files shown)


chrmod:
@stefanpenner @rwjblue please take a look. the tests could probably be less verbose so if you have suggestion on how to improve please share.

lib/cli.js
```diff
@@ -121,6 +123,25 @@ function getWatcher(isWatching) {
...
+  const watcher = options.watcher;
+
+  if (watcher && ['polling', 'watchman', 'node', 'events'].indexOf(watcher) === -1) {
+    throw new Error(`Unknown watcher type --watcher=[polling|watchman|node] but was: ${watcher}`);
+  }
+
+  return {
+    saneOptions: {
+      poll: watcher === 'polling',
+      watchman: watcher === 'watchman' || watcher === 'events' || !watcher,
```
stefanpenner: I think we may want to consider dropping `events` as `watchman` is both clear and concise and already covers that one.

stefanpenner: If there is no watcher specified,  it likely isn't safe to just assume `watchman` (that would also be backwards incompatible).

We should likely either default to `node` or use `execSync` to see if the `watchman version` returns some minimum version.

chrmod: seems also ci detected the problem. fixing it now

chrmod:
@stefanpenner I've changed default back to `node` and tests are passing again

stefanpenner:
@chrmod LGTM.

* I think future PR could try to autodetect watchman and prefer that
* I plan to get a broccoli release out this weekend, so we can get your improvements in the wild

chrmod:
@stefanpenner great, thanks for pushing it forward

PR #346, opened 2017-12-09
cli: detects watchman

chrmod:
as suggested in https://github.com/broccolijs/broccoli/pull/345#issuecomment-350309685

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -2,7 +2,9 @@
 
 const RSVP = require('rsvp');
 const TreeSync = require('tree-sync');
+const childProcess = require('child_process');
 const fs = require('fs');
+const semver = require('semver');
 
 const broccoli = require('./index');
 const messages = require('./messages');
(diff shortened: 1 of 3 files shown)


lib/cli.js
```diff
@@ -127,17 +128,27 @@ function buildWatcherOptions(options) {
...
 
   if (watcher && ['polling', 'watchman', 'node', 'events'].indexOf(watcher) === -1) {
     throw new Error(`Unknown watcher type --watcher=[polling|watchman|node] but was: ${watcher}`);
   }
 
+  if (!watcher) {
+    try {
+      childProcess.execSync('watchman version');
+      // if `watchman version` did not throw, the watchman is installed and can be used
+      watcher = 'watchman';
```
stefanpenner: A little insider knowledge, I believe we should only support watchman `>= 3.7` as it has a feature which automatically shares common watches if possible.

This version has also been around for quite some time, so it should be perfectly reasonable for us to do that.

```
let version = execSync('watchman version');
semver.satisfies(version, '>= 3.7.0'),
```

---

For reference, ember-cli has the following: https://github.com/ember-cli/ember-cli/blob/master/lib/models/watch-detector.js#L197 Which tests many failure scenarios and tries to choose the best possible watcher.

I would not be apposed to us as a next  step, extracting that into a common module which can be used in both projects. As it has been pretty thoroughly battle tested, and comes with some handy tests: https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/watch-detector-test.js

If extracted, I could put it up at `ember-cli/watch-detector`, thoughts?

chrmod: makes complete sense. lets do that. I will prepare the project and when ready we can move it under ember-cli organization

stefanpenner: :+1:

chrmod:
@stefanpenner have updated this PR to check for watchman version

lib/cli.js
```diff
@@ -127,17 +129,31 @@ function buildWatcherOptions(options) {
...
   if (watcher && ['polling', 'watchman', 'node', 'events'].indexOf(watcher) === -1) {
     throw new Error(`Unknown watcher type --watcher=[polling|watchman|node] but was: ${watcher}`);
   }
 
+  if (!watcher) {
+    try {
+      const output = childProcess.execSync('watchman version');
+      const version = JSON.parse(output).version;
+      if (!semver.satisfies(version, '>= 3.7.0')) {
+        throw new Error('watchman version older than 3.7.0');
```
stefanpenner: When auto-detecting, it may not be a good idea to error. Rather it should be considered when deciding `node` vs `watchman`. The reason is, other executables can be named `watchman`... there is exists one of these on the npm registry...

I think this will all be correctly handled once we extract watch-detector.

chrmod:
closing in favor of https://github.com/broccolijs/broccoli/pull/351

PR #347, opened 2017-12-11
Make tests more resilient (don’t crash if an obviously used port is b…

stefanpenner:
…eing used)

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "mocha": "^4.0.1",
     "mocha-eslint": "^4.1.0",
     "multidep": "^2.0.2",
+    "portfinder": "^1.0.13",
     "prettier": "^1.8.2",
     "semver": "^5.4.1",
     "sinon": "^4.1.2",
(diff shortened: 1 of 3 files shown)


chrmod:
@stefanpenner I've noticed that if tests fail due to taken port it often means the previous test did not finish/cleanup correctly. It sometimes help to detect a real problem with test test or with the code, so I'm not sure if this change is worth it. 
Also, there are some cli tests that also start the server, so if you decide to go with this direction it may be worth take a look on other ones. (one way to detect which ones start serve is by looking on console output).

stefanpenner:
> @stefanpenner I've noticed that if tests fail due to taken port it often means the previous test did not finish/cleanup correctly.

This will still happen, as the scenario where this is possible is during the same test run, and we only use 1 `PORT` per test run.

> It sometimes help to detect a real problem with test test or with the code, so I'm not sure if this change is worth it.

I believe the only way cross test run leaks can happen is if we use subprocesses for anything, and we do not.

stefanpenner:
> Also, there are some cli tests that also start the server,
> Good catch, I will update these.

Actually I do not believe they do. As the mocks don't actually call into the true server stuff.

PR #350, opened 2017-12-12
Add Build Cancellation

stefanpenner:
### Descrption

For larger builds, it is important that a mid-build `<ctrl-c>` can both gracefully exit, but also exist as fast as possible. In-order to accomplish this, we need to be able to terminate the build after the current step completes.

Also for larger builds, it is important that if a new change event occurs mid-build, that we can pause the build after the current step, and resume it from the beginning (in a safe way). This provides a good user-experience, as 2 changes slower then the throttle threshold but faster then a rebuild will still only produce 1 consistent output. No accidentally mid-build stale outputs.

```
builder.cancel().then(() => /* the build was cancelled */);
```

### TODOS

- [x] a fulfilled `cancel` guarantees to leave the build in a resumable state
- [x]  a rejected `cancel` suggests a unrecoverable build
- [x] `builder.cancel()` will currently wait for the current step to an, before cancelling the build. This is to ensure the build is resumable, and can be lifted in the future with more work.
- [x] a `builder.buid()` is not allowed if the `builder` currently has a pending build or cancellation
- [x] `builder.cancel` is idempotent
- [x] events (‘nodeStart’ ‘nodeEnd’) remain balanced in-light or cancellation
- [x] CancellationErrors are `isSilent` and `isCancellation` to play nicely with consumers like ember-cli

TODO: 

- [ ] mid-build change events should cancel the current build and queue up behind the existing cancel before kicking off another build.
- [x] better error messages
- [ ] wire up signals so the CLI gets cancellation on <ctrl-c> etc.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -15,6 +15,7 @@ const SourceNodeWrapper = require('./wrappers/source-node');
 const BuilderError = require('./errors/builder');
 const NodeSetupError = require('./errors/node-setup');
 const BuildError = require('./errors/build');
+const CancelationRequest = require('./cancellation-request');
 
 // Clean up left-over temporary directories on uncaught exception.
 tmp.setGracefulCleanup();
@@ -61,6 +62,7 @@ module.exports = class Builder {
     this.checkInputPathsExist();
 
     this.setupTmpDirs();
+    this._cancelationRequest = undefined;
 
     // Now that temporary directories are set up, we need to run the rest of the
     // constructor in a try/catch block to clean them up if necessary.
@@ -81,8 +83,16 @@ module.exports = class Builder {
   // This method will never throw, and it will never be rejected with anything
   // other than a BuildError.
   build() {
+    if (this._cancelationRequest) {
+      return RSVP.Promise.reject(
+        new BuilderError('Cannot start a build if one is already running')
+      );
+    }
+
+    let promise = RSVP.Promise.resolve();
+
     this.buildId++;
-    let promise = RSVP.resolve();
+
     this.nodeWrappers.forEach(nw => {
       // We use `.forEach` instead of `for` to close nested functions over `nw`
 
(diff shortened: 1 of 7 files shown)


lib/builder.js
```diff
@@ -81,8 +83,14 @@ module.exports = class Builder {
   // This method will never throw, and it will never be rejected with anything
   // other than a BuildError.
   build() {
+    if (this._cancelationRequest) {
+      return RSVP.Promise.reject('WTF');
```
stefanpenner: better error message, likely a `BuildError` with a message similar to`cannot build mid-build`

lib/cancellation-request.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const CancelationError = require('./errors/cancelation');
+
+module.exports = class CancelationRequest {
```
stefanpenner: this class needs to be unit tested

lib/errors/build.js
```diff
@@ -44,7 +44,10 @@ module.exports = class BuildError extends BuilderError {
       instantiationStack;
 
     super(message);
+    // consider for x in y
```
stefanpenner: should we?

lib/errors/cancelation.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+// Base class for builder errors
+module.exports = class Cancellation extends Error {
```
stefanpenner: this class needs to be unit tested

test/builder_test.js
```diff
@@ -29,6 +29,9 @@ RSVP.on('error', error => {
   throw error;
 });
 
+function wait(time) {
```
rwjblue: I like calling this `settled` because `await wait()` is weird 😝

test/builder_test.js
```diff
@@ -718,4 +721,167 @@ describe('Builder', function() {
...
+      const pipeline = new Builder(stepB);
+      let stepAIsComplete = false;
+      let cancellingIsComplete = false;
+
+      const building = expect(pipeline.build()).to.eventually.be.rejectedWith('I is CANCELLED');
+
+      let resolveCancel;
+      const cancelling = new RSVP.Promise(resolve => (resolveCancel = resolve));
+
+      stepA.build = function() {
```
stefanpenner: subclass rather then hackity hack

test/builder_test.js
stefanpenner: i think this is more `nextTick`

oligriffiths:
@stefanpenner what’s the state of play with this? Looks pretty good rn

rwjblue:
@oligriffiths / @stefanpenner - Lets move the remaining TODO's:

> * mid-build change events should cancel the current build and queue up behind the existing cancel before kicking off another build.
> * wire up signals so the CLI gets cancellation on etc.

PR #354, opened 2018-01-29
Overwrite output

oligriffiths:
This PR adds an `--overwrite` CLI option to the build command.
It's very frustrating to have to remove the output directory before each build, so this fixes that issue.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -5,6 +5,7 @@ const TreeSync = require('tree-sync');
 const childProcess = require('child_process');
 const fs = require('fs');
 const WatchDetector = require('watch-detector');
+const rimraf = require('rimraf');
 
 const broccoli = require('./index');
 const messages = require('./messages');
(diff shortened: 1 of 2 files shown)


lib/cli.js
```diff
@@ -54,6 +55,7 @@ module.exports = function broccoliCLI(args) {
     .option('--output-path <path>', 'the path to target output folder')
     .option('--watch', 'turn on the watcher')
     .option('--watcher <watcher>', 'select sane watcher mode')
+    .option('--overwrite', 'overwrite the <target> directory')
```
chrmod: should probably refer to `<target>` by `output-path` name

oligriffiths:
This will still prevent accidental deletions as it’s an optional flag. And I’m any case, the build should be deterministic, so you can run multiple times and the output is the same if the input doesn’t change.

oligriffiths:
@chemod I like the idea of doing this automatically inside the project folder

chrmod:
@oligriffiths one topic to consider is when broccoli is used from some script that inject output path from something like env variable. The output folder is protected from accidental deletion in case the passed value was unexpectedly change. Is such cases developer probably controls the script already, so you can run `rm -rf` on such variable.

Another alternative I see is to have .broccolirc file with a default params, so we can skip the guard.

oligriffiths:
Also updated the PR so lint & tests pass

oligriffiths:
@chrmod thoughts?

oligriffiths:
bump

oligriffiths:
I see in ember-cli that there is protection to ensure the directory to delete isn’t a parent of the project. Should probably add that.

rwjblue:
Ya, good idea!

rwjblue:
@oligriffiths - Wanna whip that up as a separate PR?

oligriffiths:
Sure thinf

PR #358, opened 2018-03-20
Add overwrite protection to parent directory

oligriffiths:
This PR ensures that if using the `--overwrite` option and `target` is a parent directory of the project, the process will exit(1).
This is inline with ember-cli.
Also adds some more tests for `--overwrite` option.

Question, why do we even allow the `target` directory to be a literal parent of the project, I don't really see why you would want that?

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -5,6 +5,7 @@ const TreeSync = require('tree-sync');
 const childProcess = require('child_process');
 const fs = require('fs');
 const WatchDetector = require('watch-detector');
+const path = require('path');
 
 const broccoli = require('./index');
 const messages = require('./messages');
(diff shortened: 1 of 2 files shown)


oligriffiths:
hmm, tests failing but not locally

oligriffiths:
@rwjblue do you know of any reason to allow `[target]` to be a direct parent directory to the build file?

test/cli_test.js
```diff
@@ -36,6 +36,9 @@ describe('cli', function() {
   });
 
   describe('build', function() {
+    beforeEach(function() {
+      rimraf.sync('dist');
```
rwjblue: We should generally not have to cleanup both before and after like this. Having to cleanup before generally indicates some other test is leaving us in an invalid starting state...

oligriffiths:
I think I found the problem. 
Was talking with Toby, but it seems that when mocking process.exit(), statements *after* the exit line are run in unit tests because the process does NOT exit and continues to run. With no return statement this means a build can run when an error is produced. I will whip up a quick test of this theory but this is what I’m seeing in existing tests.

lib/cli.js
```diff
@@ -157,17 +157,36 @@ function buildWatcherOptions(options) {
...
       outputDir +
         '/ already exists; we cannot build into an existing directory, ' +
-        'pass --overwrite to auto-delete the output directory'
+        'pass --overwrite to overwrite the output directory'
     );
     process.exit(1);
   }
 }
+
+function isParentDirectory(outputPath) {
```
chrmod: maybe you can use `findup-sync` to detect project root by locating `package.json` ?

oligriffiths: will check

lib/cli.js
```diff
@@ -177,16 +177,36 @@ function buildWatcherOptions(options) {
   };
 }
 
-function guardOutputDir(outputDir, removeExisting) {
-  if (fs.existsSync(outputDir)) {
-    if (removeExisting) {
-      rimraf.sync(outputDir);
```
oligriffiths: Removing rimraf as broccoli already handles doing a diff of the output directory and the new files to be added/removed, so rimraf actually makes it less efficient. https://github.com/broccolijs/broccoli/blob/master/lib/cli.js#L104

oligriffiths:
@rwjblue @chrmod updated, tests passing.

rwjblue:
Good work, thank you!!

PR #359, opened 2018-03-27
 Add CliError & returns to exits

oligriffiths:
Whilst working on https://github.com/broccolijs/broccoli/pull/358 I've discovered that when mocking `process.exit()`, execution continues after the `exit()` which causes weird test behavior, as the process isn't actually exiting.

This PR adds some `return` lines in before each `process.exit()` to ensure that we stop processing.
It also adds a `CliError` class, so the `guardOutputDir()` method can throw an exception that can be caught, and the process.exit can be contained within the main cli program, and thus returned.

As such, a couple of tests needed to be updated as they were using APIs as a result of the execution not actually exiting. They are now synchronous rather than async with a promise.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -135,7 +135,7 @@ module.exports = function broccoliCLI(args) {
 
   if (!actionPromise) {
     program.outputHelp();
-    process.exit(1);
+    return process.exit(1);
   }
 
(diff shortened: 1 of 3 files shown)


stefanpenner:
good catch!

PR #361, opened 2018-04-03
Remove test/.jshintrc and replace with test/.eslintrc.js

oligriffiths:
We don't use jshint any more, looks like this was never changed.
Not sure what `expr` is or what the `eslint` equivalent is

### DIFF
--- a/test/.eslintrc.js
+++ b/test/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  extends: "../.eslintrc.js",
+  env: {
+    mocha: true
+  }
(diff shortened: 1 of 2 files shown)


test/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: "../.eslintrc.js",
```
rwjblue: I don't think you need this, eslint extends the parent config by default

oligriffiths: Cool, done

PR #362, opened 2018-04-29
Add Heimdall support

oligriffiths:
This PR adds heimdall stats support to Broccoli master.

This is ported from https://github.com/ember-cli/broccoli-builder
I've tried to setup the test in as close a representation to the original as possible.

What we're doing is registering the `beginNode` and `endNode` events to start/stop the heimdall timer, but these events are no longer "nested" as they are in Broccoli < 1.0. As a result, to get the nested structure that heimdall needs to do flame graphs, etc, we need to reconstruct this graph using the Broccoli node graph.

Additionally, because Broccoli re-uses nodes (for source directories), and heimdall does NOT allow you to have a node with multiple parents, we're simulating a new heimdall node, in the same way as we're doing it in Broccoli < 1.0 https://github.com/ember-cli/broccoli-builder/blob/0-18-x/lib/builder.js#L101-L107 => https://github.com/broccolijs/broccoli/pull/362/files#diff-dde50c51d91de11d407f53423f3baeecR399]]]

Here's a vis output (running BROCCOLI_VIZ=1 ember build) for `broccoli 1.0` and current `master` for 2 builds:

I used https://github.com/ember-cli/broccoli-builder/pull/27 for the `broccoli-builder` branch as this makes the labels more informative and closer to `broccoli` (which I think is good).

`full` - a regular build of ember on a fresh application
`slim` - a simple build (2 nodes and a merge)

[Archive.zip](https://github.com/broccolijs/broccoli/files/2020799/Archive.zip)

These can be uploaded to https://rwjblue.github.io/heimdalljs-visualizer

They produce the following results:

`broccoli-full`:

image
image

`master-full`:
image

image

`broccoli-slim`:

image

image

`master-slim`:

image

image

As you can see, both flame graphs are the same, and `broccoli-1` is slightly faster

# Note
There appears to be a bug in the ordering of the output in the flame graph, looking into it...

### DIFF
--- a/test/builder_test.js
+++ b/test/builder_test.js
@@ -19,6 +19,7 @@ const sinonChai = require('sinon-chai');
 chai.use(sinonChai);
 const multidepRequire = require('multidep')('test/multidep.json');
 const semver = require('semver');
+const heimdall = require('heimdalljs');
 
 const Plugin = multidepRequire('broccoli-plugin', '1.3.0');
 const broccoliSource = multidepRequire('broccoli-source', '1.1.0');
@@ -60,6 +61,10 @@ describe('Builder', function() {
   }
 
   let builder;
+
+  beforeEach(function() {
+    heimdall._reset();
+  });
 
   afterEach(function() {
     sinon.restore();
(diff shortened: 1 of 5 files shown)


test/builder_test.js
```diff
@@ -895,4 +900,113 @@ describe('Builder', function() {
         });
     });
   });
+
+  describe('heimdall stats', function() {
+    it('produces stats', function(done) {
```
stefanpenner: let use mocha's built in promise support, rather then `done`. This way, error handling is handled for us by default in tests.

This looks like:

```
it('foo', function() {
  return promise.then(() => {});
})
```

oligriffiths:
@stefanpenner any idea how we can get the nested data for heidmall. Seems like it needs to be:

start(a)
start(b)
stop(b)
stop(a)

In order to make B a child of A. 

Currently the node event system does 

Start a
Stop a
Start b
Stop b

stefanpenner:
broccoli 1.0 no longer evaluates nodes recursively. Rather it essentially top-sorts the plugin graph ahead of time, and produces a linear array of nodeWrappers, related code:

https://github.com/broccolijs/broccoli/blob/7d97ebaa55dd81d73b73b4d663ce66acdf9cb714/lib/builder.js#L144-L146 


This `nodeWrappers` array is traversed, emitting the events between invocations of `build`, not as a parent of the recursive pattern from before.

What I am saying, is. The new pattern in broccoli is much nicer, but it does lose the parent/child hierarchy we had before, and thus may require a difference.

What likely could be done, is to emit as we do now. But ensure the `start` of a node, includes an array of ids for its inputs. Thoughts?

oligriffiths:
Yes that was the conclusion I came to. I think adding the parent ids to the entry and reconstructing the recursive output would be a good idea. I also notice that in the broccoli-builder implementation, the total time for each node is the self time plus the sum of the inputs, which we should also preserve. This would mean iterating the heidmall tree from top down. I can work this out.

oligriffiths:
@stefanpenner Super. Thanks

stefanpenner:
Its a bummer we have to do this, but as part of this step I don't see a good option. I do believe as part of the 1x series, we will re-invest in heimdall, to make it and broccoli better friends.

krisselden:
I actually find the current thing to be confusing because it makes the first parent look responsible for time it wasn't responsible for.

oligriffiths:
@krisselden yeah I agreee. Heimdall is not able to properly model the broccoli graph. The fact that broccoli-builder would make “fake” nodes in order to satisfy shared plugins shows that the model is incorrect. 

Unfortunately heimdall would require a decent rewrite to support this, or something else entirely.

krisselden:
Not sure why it would model the broccoli graph, Id rather put inputBroccoli ids alongside broccoliId. Heimdal represnts what you’d put into a flame chart.

I hate that the visualizer now overweights the consuming node, it’s misleading.

oligriffiths:
@krisselden a flame graph can’t accurately model a broccoli tree due to the fact that nodes are shared between inputs. Source nodes for example are reused. Even broccoli-builder had this same issue.

PR #363, opened 2018-04-30
 Return outputNode from build promise

oligriffiths:
This PR adds 1 feature to enable broccoli master to work in ember-cli (plus a small tweak to re-implement the `summarize` method below):

Make the `build` promise return `outputNode` as the result, this is consistent with `broccoli-builder` and required for `ember-cli` to function correctly. This can be seen here:
https://github.com/ember-cli/broccoli-builder/blob/d3add6436f8e1c29f4532482bf7fda8dc10b0017/lib/builder.js#L52
The above `readAndReturnNodeFor` returns the last node, the `outputNode` in broccoli nomenclature.
This is then passed to `summarize`:
https://github.com/ember-cli/broccoli-builder/blob/d3add6436f8e1c29f4532482bf7fda8dc10b0017/lib/builder.js#L54
which reformats it into an object:
https://github.com/ember-cli/broccoli-builder/blob/d3add6436f8e1c29f4532482bf7fda8dc10b0017/lib/builder.js#L29-L34
This object is then passed from the ember-cli chain of `thens`:
https://github.com/ember-cli/ember-cli/blob/689bdbfee75ef92a33a86466669062bfb7507531/lib/models/builder.js#L156
To `processBuildResult` which uses the `.directory` property:
https://github.com/ember-cli/ember-cli/blob/689bdbfee75ef92a33a86466669062bfb7507531/lib/models/builder.js#L113

Once these are merged, this makes broccoli work in ember-cli https://github.com/ember-cli/ember-cli/pull/7798

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -121,9 +121,13 @@ module.exports = class Builder {
 
     this._cancelationRequest = new CancelationRequest(promise);
 
-    return promise.finally(() => {
-      this._cancelationRequest = null;
-    });
+    return promise
+      .then(() => {
+        return this.outputNodeWrapper;
+      })
+      .finally(() => {
+        this._cancelationRequest = null;
+      });
   }
 
   cancel() {
--- a/test/builder_test.js
+++ b/test/builder_test.js
@@ -6,6 +6,7 @@ const path = require('path');
 const RSVP = require('rsvp');
 const tmp = require('tmp');
 const broccoli = require('..');
+const TransformNode = require('../lib/wrappers/transform-node');
 const makePlugins = require('./plugins');
 const Builder = broccoli.Builder;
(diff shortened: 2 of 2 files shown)


oligriffiths:
Ah yes of course

oligriffiths:
@stefanpenner updated per comments. Please see https://github.com/ember-cli/ember-cli/pull/7798 about whether we should return `outputNodeWrapper` or `outputNodeWrapper.node`...

oligriffiths:
I don't know why this is failing on some node versions, it passes on my mac. Can anyone shed some light?

oligriffiths:
So I think returning `nodeWrapper` is the right thing as this contains the `id` property that I need for the heimdall stuff

PR #366, opened 2018-06-11
Update dependencies

krisselden:


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
rwjblue: In general, editor specific config should go in a global gitignore

PR #367, opened 2018-06-11
Remove RSVP

krisselden:


rwjblue:
I think this needs a rebase now that I merged the dependency bump PR...

lib/builder.js
```diff
@@ -104,16 +104,18 @@ module.exports = class Builder {
...
+            .then(() => this._cancelationRequest.throwIfRequested())
+            .then(() => this.emit('beginNode', nw))
+            .then(() => nw.build()),
+          () => {
             if (this._cancelationRequest.isCancelled) {
               return;
             }
-            this.trigger('endNode', nw);
-          })
+            this.emit('endNode', nw);
```
stefanpenner: don't we need to omit this in the finally clause?

Issue #368, opened 2018-06-12
readme should also describe programmatic builder usage


stefanpenner:
something like: https://github.com/broccolijs/broccoli/blob/master/lib/builder.js#L29-L40

cafreeman:
I'd like to take a stab at this in the next few days

cafreeman:
@stefanpenner just to close the loop, I opened a PR for this [here](https://github.com/broccolijs/broccoli/pull/369).

stefanpenner:
@cafreeman ah fantastic, looks great!

PR #369, opened 2018-06-23
Describe programmatic Broccoli Builder usage in README

cafreeman:
This PR attempts to address https://github.com/broccolijs/broccoli/issues/368.

Let me know if I've gotten anything wrong, or if there are issues with the style and I'm happy to address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,68 @@ running `broccoli build the-output` would generate the following folder:
 
 You can find plugins under the [broccoli-plugin keyword](https://www.npmjs.org/browse/keyword/broccoli-plugin) on npm.
 
+## Using Broccoli Programmatically
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,68 @@ running `broccoli build the-output` would generate the following folder:
...
+
+const outputDir = 'dist';
+const outputTree = new TreeSync(builder.outputPath, outputDir);
+
+builder.build()
+  .then(() => {
+    // Calling `sync` will synchronize the contents of the builder's `outPath` with our output directory.
+    outputTree.sync();
+  })
+  .finally(() => {
```
krisselden: finally does not exist until node 10

rwjblue: yes, but you would need to do it in both success and failure cases (e.g. `.then` + `.catch` or `.then(cb, cb)`)

cafreeman: Gotcha. Will do!

cafreeman: @rwjblue updated!

PR #371, opened 2018-07-23
Swap mime package for mime-types

oligriffiths:
# What
Swap mime package for mime-types

# Why
I noticed that the latest master was hanging on `broccoli serve` and never loading. I dug into it and found the issue to be the `mime` package.
Since we changed to `mime-types` in https://github.com/ember-cli/broccoli-middleware/pull/43 it seems fitting to do it here, and fixes the issue.

# Test
`broccoli serve` should load correctly.


rwjblue:
We also really need a test for this :( the fact that `broccoli serve` was busted and all of our tests were happy makes me 😢

oligriffiths:
@rwjblue 
> We should migrate to mime-types
Ok, how do you want to do this then?
- Merge this to fix the issue and work on replacing `mime` with `mime-types`
- Revert the package to v 1.5 and work on replacing `mime` with `mime-types`
- Work on replacing `mime` with `mime-types`

> We also really need a test for this
100% agree. Not sure how best to test serve, or how one would do a curl or something in tests.

krisselden:
You need to do more than just switch it in the package json

oligriffiths:
@krisselden done

test/server_test.js
```diff
@@ -105,4 +108,42 @@ describe('server', function() {
...
+      server.onBuildSuccessful = function() {
+        try {
+          onBuildSuccessful();
+          resolve();
+        } catch (e) {
+          reject(e);
+        }
+      };
+    }).then(() =>
+      promiseFinally(
```
stefanpenner: this seems funky

krisselden: this is gone

lib/watcher_adapter.js
```diff
@@ -27,43 +27,40 @@ class WatcherAdapter extends EventEmitter {
...
         watcher.on('error', err => {
           logger.debug('error', err);
           this.emit('error', err);
         });
-
-        watcher.on('ready', () => {
-          logger.debug('ready', watchedPath);
-          resolve(watcher);
-        });
+        logger.debug('ready', watchedPath);
```
rwjblue: We only want the logger.debug for the first watcher ready event?

lib/watcher_adapter.js
```diff
@@ -27,43 +27,40 @@ class WatcherAdapter extends EventEmitter {
...
+          watcher.close(err => {
+            if (err) reject(err);
+            else resolve();
+          })
+        )
+    );
     this.watchers.length = 0;
-
-    return Promise.all(closing).then(function() {});
+    return Promise.all(closing).then(() => {});
```
rwjblue: Just curious, why do we need the empty function here?

lib/watcher_adapter.js
krisselden: why? do we not want to know all the watched paths?

lib/watcher_adapter.js
krisselden: There is a test for `Promise<undefined>` vs `Promise<undefined[]>` not sure why it wasn't failing before.  I believe there are still issues with the promises, I would like to convert to typescript so we can use the tslint no-floating-promises

PR #379, opened 2018-10-24
Fix issue https://github.com/ember-cli/ember-cli/issues/8118

krisselden:
I'm not a fan of doing this here, it isn't actually representing what happens.

I believe we should output the input ids on the node so you could build this in a visualizer or post process.

Fixes https://github.com/ember-cli/ember-cli/issues/8118

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -360,6 +360,8 @@ module.exports = class Builder extends EventEmitter {
         label: node.label,
         broccoliNode: true,
         broccoliId: node.id,
+        // we should do this instead of reparentNodes
+        // broccoliInputIds: node.inputNodeWrappers.map(input => input.id),
         broccoliCachedNode: false,
         broccoliPluginName: node.nodeInfo.name,
       });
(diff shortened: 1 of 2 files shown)


krisselden:
This fixes the issue on ember cli but I have many questions after doing it.

A) I dislike rewriting what actually happened my preference would be to add the input ids alongside the rest of the broccoli info.

B) why do we compute total? Can’t the consumer do that?

The cause of this was traversing “cached node inputs” which is bounded but would overflow a recursive algorithm easier.  

Removing the recursion here just fails as soon as you try BROCCOLI_VIZ=1  inside of toJSON() in heimdall which is recursive.  So I made it stop on “cached” nodes which are 0 and you can’t even see in the visualizer.

So this is a long way of saying while this fixes the issue I disagree with what it is doing.

oligriffiths:
@krisselden the reason for doing this is to replicate what broccoli-builder was doing (as it worked recursively when calling plugins) vs broccoli 2 that does each node separately and non-recursively.

The reason that heimdall is not able to accurately represent the broccoli node graph for timing purposes is because of shared source nodes. Flame graphs work on the basis of each element in the graph being unique instances of time, with Broccoli that's not the case.

Say we have the following broccoli tree:
```
        app
     /      \
filter(js)  filter(scss)
    |        |
babel        scss
     \      /
      merge
```
As `app` is shared how would one attribute its time in a flame graph?

Broccoli-builder suffered the same issue, hence why it duplicated nodes that had already been processed.

rwjblue:
Thank you for fixing up that windows failure @krisselden.

oligriffiths:
Thanks @krisselden

PR #380, opened 2018-10-25
Heimdall fix recursive calls

oligriffiths:
This PR branches off the work @krisselden did to change from recursive to a stack https://github.com/broccolijs/broccoli/pull/379

I concluded the main difference was *not* recursing for nodes already processed, the fix for that can be seen here https://github.com/broccolijs/broccoli/commit/3fb21cad1d228081892ba99c826285eb8ad49c30

I also cleaned up some unnecessary work of indexing heimdall nodes by broccoli ID as each ndoeWrapper has a __heimdall__ property with the heimdall node anyway. Also used a set for nodes that have been seen already.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -375,26 +375,11 @@ module.exports = class Builder extends EventEmitter {
 
   buildHeimdallTree(outputNodeWrapper) {
     const heimdallRootNode = outputNodeWrapper.__heimdall__;
-    const heimdallByBroccoliId = {};
-
     if (!heimdallRootNode) {
       return;
     }
 
-    heimdallRootNode.parent.forEachChild(child => {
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm, it is not clear to me why this is better than #379.

oligriffiths:
@rwjblue it's fewer changes and much less code...

rwjblue:
Maybe but I don’t think that we are optimizing for LOC or reduced diffs though are we?

oligriffiths:
@rwjblue AFAICT, the only difference between mine and and kris's is his works iteratively and mine works recursively. Am I missing something else? AFAIK more code to achieve the same thing isn't usually preferred

rwjblue:
Recursing will blow the stack for large graphs

PR #384, opened 2018-12-06
Make --overwrite the default but protect writing to direct parents

oligriffiths:
After discussion with @stefanpenner we decided it makes sense to have `--overwrite` option be the default. Other build tools will rebuild to an existing output directory. Given that we write to `dist` by default, having to `rm -r dist` ever time you run build is annoying and unnecessary.

An optional middle ground would be to only protect the output directory when a target outside the cwd is also specified, but I'm not sure that's necessary either really, however it would save doing `--output-target=$HOME/Desktop` from blowing away your desktop for example...


oligriffiths:
@rwjblue mind taking a look at this?

PR #385, opened 2019-02-14
Add ES6 Module support

oligriffiths:
This PR adds ESM support to the Brocfile.js
You can now use `import/export` etc in your Brocfile and in plugins (though we will need to document this so they're backwards compatible).

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,16 +92,17 @@ The following `Brocfile.js` exports the `app/` subdirectory as `appkit/`:
 
 ```js
 // Brocfile.js
-import funnel from 'broccoli-funnel';
+import Funnel from 'broccoli-funnel';
 
-export default () => funnel('app', {
+export default () => new Funnel('app', {
   destDir: 'appkit'
  })
 }
 ```
 
-Broccoli support ESM modules via [esm](https://www.npmjs.com/package/esm). You can also use regular CommonJS `require`
-and `module.exports` if you prefer.
+Broccoli supports [ES6 modules](https://nodejs.org/api/esm.html) via [esm](https://www.npmjs.com/package/esm).
+You can also use regular CommonJS `require` and `module.exports` if you prefer, however ESM is the future of Node, and
+the recommended syntax to use.
 
 That example uses the plugin
 [`broccoli-funnel`](https://www.npmjs.com/package/broccoli-funnel).
@@ -130,32 +131,32 @@ By way of example, let's assume we have a graph of Broccoli nodes constructed vi
 
 ```js
 // non Brocfile.js, regular commonjs
-const funnel = require('broccoli-funnel');
-const merge = require('broccoli-merge');
+const Funnel = require('broccoli-funnel');
+const Merge = require('broccoli-merge');
 
-const html = funnel(appRoot, {
+const html = new Funnel(appRoot, {
   files: ['index.html'],
   annotation: 'Index file'
 })
 
-const js = funnel(appRoot, {
+const js = new Funnel(appRoot, {
   files: ['app.js'],
   destDir: '/assets',
   annotation: 'JS Files'
 });
 
-const css = funnel(appRoot, {
+const css = new Funnel(appRoot, {
   srcDir: 'styles',
   files: ['app.css'],
   destDir: '/assets',
   annotation: 'CSS Files'
 });
 
-const public = funnel(appRoot, {
+const public = new Funnel(appRoot, {
   annotation: 'Public Files'
 });
 
-const tree = merge([html, js, css, public]);
+const tree = new Merge([html, js, css, public]);
 ```
 
 At this point, `tree` is a graph of nodes, each of which can represent either an input or a transformation that we want to perform. In other words, `tree` is an abstract set of operations, *not* a concrete set of output files.
@@ -169,9 +170,9 @@ Since we typically want do more than write to a temporary folder, we'll also use
 ```js
 const { Builder } = require('broccoli');
 const TreeSync = require('tree-sync');
-const merge = require('broccoli-merge');
+const Merge = require('broccoli-merge');
 // ...snip...
-const tree = merge([html, js, css, public]);
+const tree = new Merge([html, js, css, public]);
 
 const builder = new Builder(tree);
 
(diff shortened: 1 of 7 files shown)


stefanpenner:
Awesome


I wonder if we should stay closer to node, and  only support this new syntax for Brocfile.mjs files and leave Brocfile.js in classic cjs mode.

Cc @jdalton

jdalton:
👋 @stefanpenner!

Allowing `.js` for ES modules is great! Node's experimental ES module support can, and will, change. Likely as soon as April. There will be support for ES module files other than `.mjs`, so a way for `.js` files to be treated as ES modules. However, the level of interop with CJS is still not finalized.

oligriffiths:
> I wonder if we should stay closer to node, and only support this new syntax for Brocfile.mjs files and leave Brocfile.js in classic cjs mode.

@stefanpenner if Node plans to support .js files containing ESM, then I'd say we should do the same. Is there a downside of adding ESM support for .js files?

oligriffiths:
it appears rn with node 11.9.0, modules only work when you include an mjs file
```
➜  node node --experimental-modules index.mjs
(node:45418) ExperimentalWarning: The ESM module loader is experimental.
foo
➜  node node --experimental-modules index.js
(node:45423) ExperimentalWarning: The ESM module loader is experimental.
/private/tmp/node/index.js:1
(function (exports, require, module, __filename, __dirname) { import foo from './foo'
                                                                     ^^^

SyntaxError: Unexpected identifier
    at new Script (vm.js:84:7)
    at createScript (vm.js:264:10)
    at Proxy.runInThisContext (vm.js:312:10)
    at Module._compile (internal/modules/cjs/loader.js:694:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
    at Function.Module._load (internal/modules/cjs/loader.js:558:3)
    at createDynamicModule (internal/modules/esm/translators.js:73:15)
    at Object.meta.done (internal/modules/esm/create_dynamic_module.js:40:9)
```

stefanpenner:
@jdalton should we make both js and mid work via esm today, or should we wait re: es in us till the cjs interop is complete?

jdalton:
I'm a big fan of making things _"just work"_ today. It's why I created `esm` 😉

oligriffiths:
Yeah I feel like having it “just work” is good. We already have “just work” in ember with using modules and not requiring an mjs file. @stefanpenner Thoughts?

stefanpenner:
Sounds good to me. I just wanted to double check with @jdalton as I know he’s wired into this

lib/load_brocfile.js
```diff
@@ -29,7 +30,12 @@ module.exports = function loadBrocfile(options) {
   process.chdir(baseDir);
 
   // Load brocfile
-  const brocfile = require(brocfilePath);
+  let brocfile = esmRequire(brocfilePath);
+
+  // Esm imports require an "export default" that is returned as { default: ... }
+  if (typeof brocfile === 'object' && brocfile.default) {
```
oligriffiths: @jdalton I noticed that

```js
export default “foo”;
```
when run should esm returns
```js
{
    default: “foo”
}
```
Does this make sense then?

jdalton: Yep. Remember `null` is typeof `"object"` too. There is an undocumented option tentatively called `dedefault` which will return `default` export in cases where you only export a `default` binding.
```js
const esmRequire = require('esm')(module, {
  cjs: {
    dedefault: true
  }
});
```

jdalton: Though if you're wanting to support `.mjs` too (`esm` options don't extend to `.mjs`) you might keep your existing style check. Your check could be refined to:
```js
if (typeof brocfile === 'object' && brocfile !== null) {
  const names = Object.keys(brocfile)

  if (names.length === 0 && names[0] === 'default') {
    brocfile = brocfile.default;
  }
}
```

oligriffiths: Yeah that's an option, I'd like to keep forwards compatibility with .mjs.
I'm not sure what should happen with the following:
```js
export const FOO = 'foo';
export default tree;
```
which will produce
```js
{
  "FOO": "foo",
  "default": tree
}
```
We can either error if there are additional keys or ignore them. Technically exporting a named export won't do anything as the importing side is only looking at default so we might want to just ignore them, in which case your suggested change above would miss.

oligriffiths:
@stefanpenner sounds like we have an accord on the way forwards. LMK if this is good.

lib/load_brocfile.js
rwjblue: @oligriffiths - I'd prefer to ignore them. It seems perfectly reasonable to me for folks to have other shared exports from `Brocfile.js`...

README.md
```diff
@@ -62,15 +62,18 @@ the following folder within your project folder:
...
 
-module.exports = () => {
- return new Funnel('app', {
+export default () => funnel('app', {
   destDir: 'appkit'
  })
 }
 ```
 
+Broccoli support ESM modules via [esm](https://www.npmjs.com/package/esm). You can also use regular CommonJS `require`
```
rwjblue:
```suggestion
Broccoli supports ES modules via [esm](https://www.npmjs.com/package/esm). You can also use regular CommonJS `require`
```

README.md
```diff
@@ -97,29 +100,33 @@ In addition to using Broccoli via the combination of `broccoli-cli` and a `Brocf
...
+const funnel = require('broccoli-funnel');
+const merge = require('broccoli-merge');
+
+const html = funnel(appRoot, {
   files: ['index.html'],
   annotation: 'Index file'
 })
 
-const js = new Funnel(appRoot, {
+const js = funnel(appRoot, {
```
rwjblue: These changes (away from `new Foo`) seem unrelated to ESM, lets do them in a different PR.

oligriffiths: sure thing

lib/load_brocfile.js
oligriffiths: cool, they are ignored by default

oligriffiths:
@rwjblue requested changes applied

PR #386, opened 2019-02-15
Have Brocfile.js export a function

oligriffiths:
This PR implements the basics needed for Brocfile to return a function by default, and wrap existing brocfile exports in a function for backwards compatibility.
The intention for this is to make Broccoli work the same as Ember-CLI (to avoid confusion), and it also paves the way for us to pass in a configuration object containing things like `environment` and other build-time useful info (as Ember-CLI does).

One thing to note, is https://github.com/broccolijs/broccoli/blob/b5e4c8b7bbd378ee02d40779d483fed65439918b/lib/index.js#L7-L9 would now return a function rather than a node. This obviously changes the API, but I don't know if that's an issue or not given there was no test for this else it would have failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ npm install --global broccoli-cli
 ## Brocfile.js
 
 A `Brocfile.js` file in the project root contains the build specification. It
-should export a tree.
+should export a function that returns a tree.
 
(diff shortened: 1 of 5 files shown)


oligriffiths:
Dope. Thanks guys

PR #387, opened 2019-02-18
Support --environment CLI option and Brocfile argument

oligriffiths:
This PR implements an options object being passed to the build function.


oligriffiths:
@stefanpenner not right now. I’ll add that

PR #390, opened 2019-02-24
Add Brocfile.ts TypeScript support

oligriffiths:
This PR adds support for loading a typescript Brocfile and running it through `ts-node`.
If one provides a `Brocfile.ts`, install `ts-node` and `typescript`, it will be auto picked up and run through ts-node.
An optional `tsconfig.json` can exist alongside the `Brocfile.ts`, however defaults taken from https://github.com/stefanpenner/fs-tree-diff/pull/89/files#diff-e5e546dd2eb0351f813d63d1b39dbc48 are provided

I think I've provided a type definition for the `BrocfileOptions`, but I may have gotten this wrong, plz inform.
Tests added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ and `--environment=development` respectively.
 
 For example:
 ```js
-module.exports = (options) => {
+export default (options) => {
     // tree = ... assemble tree
 
     // In production environment, minify the files
@@ -87,6 +87,30 @@ module.exports = (options) => {
     return tree;
 }
 ```
+
+### TypeScript Support
+
+A `Brocfile.ts` can be used in place of a `Brocfile.js` and Broccoli will automatically parse this
+through [ts-node](https://www.npmjs.com/package/ts-node) to provide [TypeScript](https://www.typescriptlang.org)
+support. This allows developers to leverage type information when assembling a build pipeline. By default,
+Broccoli provides [type information](lib/index.d.ts) for the [options](#options) object passed to the build function.
+
+```ts
+import { BrocfileOptions } from 'broccoli';
+
+export default (options: BrocfileOptions) => {
+  // tree = ... assemble tree
+
+  // In production environment, minify the files
+  if (options.env === 'production') {
+    tree = minify(tree);
+  }
+
+  return tree';
+};
+
+```
+
 
 ### Using plugins in a `Brocfile.js`
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ export default (options: BrocfileOptions) => {
     tree = minify(tree);
   }
 
-  return tree';
+  return tree;
 };
 
 ```
--- a/lib/load_brocfile.js
+++ b/lib/load_brocfile.js
@@ -4,23 +4,62 @@ const path = require('path');
 const findup = require('findup-sync');
 const esmRequire = require('esm')(module);
 
+/**
+ * Require a brocfile via either ESM or TypeScript
+ *
+ * @param {String} brocfilePath The path to the brocfile
+ * @returns {*}
+ */
+function requireBrocfile(brocfilePath) {
+  let brocfile;
+
+  if (brocfilePath.match(/\.ts$/)) {
+    try {
+      require.resolve('ts-node');
+    } catch (e) {
+      throw new Error(`Cannot find module 'ts-node', please install`);
+    }
+
+    try {
+      require.resolve('typescript');
+    } catch (e) {
+      throw new Error(`Cannot find module 'typescript', please install`);
+    }
+
+    // Register ts-node typescript compiler
+    require('ts-node').register(); // eslint-disable-line node/no-unpublished-require
+
+    // Load brocfile via ts-node
+    brocfile = require(brocfilePath);
+  } else {
+    // Load brocfile via esm shim
+    brocfile = esmRequire(brocfilePath);
+  }
+
+  // ESM `export default X` is represented as module.exports = { default: X }
+  if (brocfile !== null && typeof brocfile === 'object' && brocfile.hasOwnProperty('default')) {
+    brocfile = brocfile.default;
+  }
+
+  return brocfile;
+}
+
 module.exports = function loadBrocfile(options) {
   if (!options) {
     options = {};
   }
 
   let brocfilePath;
-
   if (options.brocfilePath) {
     brocfilePath = path.resolve(options.brocfilePath);
   } else {
-    brocfilePath = findup('Brocfile.js', {
+    brocfilePath = findup('Brocfile.{ts,js}', {
       nocase: true,
     });
   }
 
   if (!brocfilePath) {
-    throw new Error('Brocfile.js not found');
+    throw new Error('Brocfile.[js|ts] not found');
   }
 
   const baseDir = options.cwd || path.dirname(brocfilePath);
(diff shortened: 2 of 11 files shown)


oligriffiths:
@stefanpenner @rwjblue have you had a chance to check this out? Am I on the right path?

oligriffiths:
@chriskrycho Would you be interested in having a look at this given your work on ember-typescript?

oligriffiths:
@mike-north Would you be interested in having a look at this given your work on ember-typescript?

lib/load_brocfile.js
```diff
@@ -4,23 +4,76 @@ const path = require('path');
...
+          moduleResolution: 'node',
+          module: 'commonjs',
+          target: 'es5',
+          baseUrl: '.',
+        },
+      };
+    }
+
+    // Ensure any previous .tsx? extensions are removed
+    delete require.extensions['.ts']; // eslint-disable-line node/no-deprecated-api
```
dfreeman: If someone has already set up `ts-node` (or otherwise decided to handle `.ts` files), shouldn't we defer to that instead of blowing it away?

lib/load_brocfile.js
```diff
@@ -4,23 +4,76 @@ const path = require('path');
...
+    const configPath = findup('tsconfig.json', {
+      nocase: true,
+    });
+
+    let options;
+
+    if (configPath) {
+      options = { project: configPath };
+    } else {
+      options = {
```
dfreeman: If no `tsconfig.json` file is found (or maybe even if it is), would it make sense to use `ts-node` in `transpileOnly` mode? It's a bit faster since it skips typechecking, and you don't risk enabling config options that will break the user.

lib/load_brocfile.js
oligriffiths: yeah I realise this was just for tests, so I will add this to the tests instead

lib/load_brocfile.js
oligriffiths: honestly not sure, happy to take advice on this for recommended default (can always be changed with tsconfig)

lib/load_brocfile.js
```diff
@@ -4,23 +4,71 @@ const path = require('path');
...
+        // Define default compilerOptions
+        compilerOptions: {
```
oligriffiths: Looking for a steer on these defaults

chriskrycho: These seem reasonable; the only one I'd consider changing is `target` – I'd set it to whatever the current Node LTS supports (which IIRC is ES2015? Might already be 2016).

oligriffiths: I think it should be node 6, which is what's in the package.json as the min node version. Agree?

chriskrycho: Seems 👍 to me!

lib/load_brocfile.js
```diff
@@ -4,23 +4,71 @@ const path = require('path');
...
+ * Require a brocfile via either ESM or TypeScript
+ *
+ * @param {String} brocfilePath The path to the brocfile
+ * @returns {*}
+ */
+function requireBrocfile(brocfilePath) {
+  let brocfile;
+
+  if (brocfilePath.match(/\.ts$/)) {
+    const configPath = findup('tsconfig.json', {
```
rwjblue: Why is this needed at all?

https://github.com/TypeStrong/ts-node#loading-tsconfigjson suggests that it will find the tsconfig.json just fine on its own, no?

lib/load_brocfile.js
mike-north: typescript already has some defaults it falls back to in the absence of a `tsconfig.json`. We may want to consider remaining consistent with that, in order to avoid introducing ts-node as a dependency, and confusing those who expect "normal" behavior from config-less typescript compilation

package.json
```diff
@@ -43,6 +44,8 @@
     "sane": "^4.0.0",
     "tmp": "0.0.33",
     "tree-sync": "^1.2.2",
+    "ts-node": "^8.0.2",
+    "typescript": "^3.3.3",
```
mike-north: you'll certainly want to make this `~3.3.3`, since typescript does not follow semver, and regularly includes breaking changes in "middle-number" releases

lib/load_brocfile.js
stefanpenner: @oligriffiths thoughts?

lib/load_brocfile.js
```diff
@@ -4,23 +4,84 @@ const path = require('path');
...
+          strict: true,
+          moduleResolution: 'node',
+          module: 'commonjs',
+          target: 'ES2015', // Node 6 per package.json
+          baseUrl: '.',
+        },
+      };
+    }
+
+    try {
```
stefanpenner: @rwjblue I think this addresses your concerns, can you confirm/deny?

package.json
oligriffiths: Dependency removed

lib/load_brocfile.js
oligriffiths: Well, the point of this is to ensure that we either use the tsconfig or we use the defaults, but not both. node-ts overwrites the tsconfig options with the passed in config https://github.com/TypeStrong/ts-node/blob/2609663bf3a5b14d76c0dfe989ceba67e68d5efa/src/index.ts#L486

oligriffiths:
Ok, so here's an update.
I've removed the `ts-node` and `typescript` dependencies, and now throw an exception if they're not installed, prompting the user to install them.
However, I have some questions on how we want `ts-node` to work by default, and if we should provide default options.

1. Should type definitions be defined for imported modules? For example, doing `import merge from 'broccoli-merge-trees'`, TS complains that there's no type information for this, and it's defaulting to `any` type. `noImplicitAny: false` is required to fix this
2. How should importing from CommonJS modules work? This is affected by `esModuleInterop: true` config option if we want it to work out of the box for `import merge from 'broccoli-merge-trees'`
3. Should we allow importing a JS file from a TS file? If so, `allowJs: true` is required.
4. if `allowJs: true` is used, the tests fail when there is more than 1 test that loads a TS brocfile (a single one passes, multiple fail):
```
     TSError: ⨯ Unable to compile TypeScript:
Brocfile.ts(2,24): error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.
Brocfile.ts(2,32): error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.
Brocfile.ts(5,23): error TS2304: Cannot find name 'exports'.
Brocfile.ts(6,43): error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.
Brocfile.ts(8,1): error TS2304: Cannot find name 'exports'.
```

😞

mike-north:
> Should type definitions be defined for imported modules? For example, doing import merge from 'broccoli-merge-trees', TS complains that there's no type information for this, and it's defaulting to any type. noImplicitAny: false is required to fix this. The downside of requiring this is that no broccoli plugins currently have type definitions, so this would be painful for early adopters.

this is a pretty general issue having to do with typescript. Consumers should either define their own types locally, PR types into the dependency (through a declaration file or a JS -> TS conversion) or add a package to DefinitelyTyped

> How should importing from CommonJS modules work? This is affected by esModuleInterop: true config option if we want it to work out of the box for import merge from 'broccoli-merge-trees'

the way to do this without forcing consumers to set this flag is
```ts
// import
import foo = require('foo');

// export
export = 'foo';
```

lib/load_brocfile.js
oligriffiths: Removed the tsconfig loading, defaults to the project dir tsconfig.json anyway, and works out of the box.

lib/load_brocfile.js
oligriffiths: Removed

README.md
```diff
@@ -88,6 +88,30 @@ module.exports = (options) => {
...
+
+export default (options: BrocfileOptions) => {
+  // tree = ... assemble tree
+
+  // In production environment, minify the files
+  if (options.env === 'production') {
+    tree = minify(tree);
+  }
+
+  return tree';
```
stefanpenner: Some funky syntax here

oligriffiths:
Thank you to everyone who contributed feedback on this PR 🎉

PR #392, opened 2019-02-25
Add BROCCOLI_ENV support

oligriffiths:
Given that https://www.npmjs.com/package/broccoli-env and https://github.com/stefanpenner/broccoli-stew/blob/master/lib/env.js#L27 support this environment variable, we should proxy its setting to allow users to use `--environment` and have these packages work OOTB. Given these packages were made by Jo and Stefan, I think setting this env var is a good idea, and allows Brocfile authors to natively support `--environment` without having to refactor their Brocfile.js in order to return a function to support the `options.env` property.

Note, setting `BROCCOLI_ENV` environment variable will *NOT* set the `options.env`.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -41,8 +41,15 @@ module.exports = function broccoliCLI(args) {
         options.environment = 'development';
       }
 
-      if (process.env.BROCCOLI_ENV && options.environment) {
-        throw new CliError('You can not use BROCCOLI_ENV and --environment/--prod/--dev together');
+      // Proxy BROCCOLI_ENV to environment
+      if (process.env.BROCCOLI_ENV) {
+        if (options.environment) {
+          throw new CliError(
+            'You can not use BROCCOLI_ENV and --environment/--prod/--dev together'
+          );
+        }
+
+        options.environment = process.env.BROCCOLI_ENV;
       }
 
       // Default environment to development
@@ -112,8 +119,15 @@ module.exports = function broccoliCLI(args) {
         options.environment = 'development';
       }
 
-      if (process.env.BROCCOLI_ENV && options.environment) {
-        throw new CliError('You can not use BROCCOLI_ENV and --environment/--prod/--dev together');
+      // Proxy BROCCOLI_ENV to environment
+      if (process.env.BROCCOLI_ENV) {
+        if (options.environment) {
+          throw new CliError(
+            'You can not use BROCCOLI_ENV and --environment/--prod/--dev together'
+          );
+        }
+
+        options.environment = process.env.BROCCOLI_ENV;
       }
 
       // Default environment to development
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -41,15 +41,8 @@ module.exports = function broccoliCLI(args) {
         options.environment = 'development';
       }
 
-      // Proxy BROCCOLI_ENV to environment
-      if (process.env.BROCCOLI_ENV) {
-        if (options.environment) {
-          throw new CliError(
-            'You can not use BROCCOLI_ENV and --environment/--prod/--dev together'
-          );
-        }
-
-        options.environment = process.env.BROCCOLI_ENV;
+      if (process.env.BROCCOLI_ENV && options.environment) {
+        throw new CliError('You can not use BROCCOLI_ENV and --environment/--prod/--dev together');
       }
 
       // Default environment to development
@@ -119,15 +112,8 @@ module.exports = function broccoliCLI(args) {
         options.environment = 'development';
       }
 
-      // Proxy BROCCOLI_ENV to environment
-      if (process.env.BROCCOLI_ENV) {
-        if (options.environment) {
-          throw new CliError(
-            'You can not use BROCCOLI_ENV and --environment/--prod/--dev together'
-          );
-        }
-
-        options.environment = process.env.BROCCOLI_ENV;
+      if (process.env.BROCCOLI_ENV && options.environment) {
+        throw new CliError('You can not use BROCCOLI_ENV and --environment/--prod/--dev together');
       }
 
       // Default environment to development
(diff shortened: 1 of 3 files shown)


lib/cli.js
```diff
@@ -41,6 +41,16 @@ module.exports = function broccoliCLI(args) {
         options.environment = 'development';
       }
 
+      if (process.env.BROCCOLI_ENV && options.environment) {
+        throw new CliError('You can not use BROCCOLI_ENV and --environment/--prod/--dev together');
```
rwjblue: I don't particularly like the idea of throwing here...

One should win (probably the `--environment` flag) over the other.

oligriffiths: I'm ok with that, `--environment` is the new one, so it can take precedence

oligriffiths:
@stefanpenner I realized a bug that this would introduce, whereby `BROCCOLI_ENV` is always clobbered with the `environment` default, even if `--environment` isn't set.

I've now updated this will the following conditions:

* If neither `BROCCOLI_ENV` or `--environment` is set, `options.environment` and `BROCCOLI_ENV` = 'development'
* If `BROCCOLI_ENV` IS NOT set, and `--environment` is set, `options.environment` = `BROCCOLI_ENV`
* If `BROCCOLI_ENV` IS set and `--environment` IS NOT set, then `BROCCOLI_ENV` retains its value, but `options.env` defaults to `development`

This ensures that this PR won't affect existing users of `broccoli-env`/`broccoli-stew` as they will likely be setting `BROCCOLI_ENV` and not `--environment`.

I'm not sure if we should proxy the value of `BROCCOLI_ENV` to `options.env` if it's not set. Currently in this PR, if `BROCCOLI_ENV` is set to `production`, then `options.env` still defaults to `development`. I don't know if it should or should not reflect the value of `BROCCOLI_ENV`, my feeling is it shouldn't because `BROCCOLI_ENV` is a legacy pattern due to Broccoli not supporting `--environment`. Thoughts?

stefanpenner:
lets get @rwjblue's sign-off

lib/cli.js
```diff
@@ -158,6 +150,22 @@ module.exports = function broccoliCLI(args) {
...
+    options.environment = 'production';
+  } else if (options.dev) {
+    options.environment = 'development';
+  }
+
+  // If environment manually supplied via CLI set BROCCOLI_ENV for broccoli-stew and broccoli-env compat
+  if (options.environment) {
+    process.env.BROCCOLI_ENV = options.environment;
+  } else {
+    options.environment = 'development';
```
rwjblue: Shouldn't this default to `process.env.BROCCOLI_ENV || 'development'`?

test/cli_test.js
```diff
@@ -192,22 +193,24 @@ describe('cli', function() {
...
+      it('defaults to { env: "development" } and sets process.env.BROCCOLI_ENV = "development"', function() {
         const spy = sinon.spy(loadBrocfile());
         sinon.stub(broccoli, 'loadBrocfile').value(() => spy);
 
-        return cli(['node', 'broccoli', 'build', 'dist']).then(() =>
-          chai.expect(spy).to.be.calledWith(sinon.match.has('env', 'development'))
-        );
+        return cli(['node', 'broccoli', 'build', 'dist']).then(() => {
+          chai.expect(spy).to.be.calledWith(sinon.match.has('env', 'development'));
+          chai.expect(process.env.BROCCOLI_ENV).to.be.undefined;
```
rwjblue: I do not think this is _ever_ expected...

test/cli_test.js
```diff
@@ -236,6 +239,28 @@ describe('cli', function() {
...
         );
       });
+
+      it('leaves BROCCOLI_ENV unaltered if --environment is not set', function() {
+        const spy = sinon.spy(loadBrocfile());
+        sinon.stub(broccoli, 'loadBrocfile').value(() => spy);
+        process.env.BROCCOLI_ENV = 'production';
+
+        return cli(['node', 'broccoli', 'build', 'dist']).then(() => {
+          chai.expect(spy).to.be.calledWith(sinon.match.has('env', 'development'));
```
rwjblue: This seems very very bizarre to me.  

```
BROCCOLI_ENV=production broccoli build
```

^ should _absolutely_ use production, no?

lib/cli.js
oligriffiths: As mentioned, I don't know if we should proxy `process.env.BROCCOLI_ENV` into `options.env`.
The intention here is to ensure that `BROCCOLI_ENV` alone will continue to work with `broccoli-env/stew`, and `broccoli-env/stew` will work with `--environment`. By allowing `BROCCOLI_ENV` to flow into `options.env` we're saying `you don't need to use --environment, you can use BROCCOLI_ENV alternatively`. Is that something we want?

test/cli_test.js
oligriffiths: This test is confirming that if `--environment` is not set, then `BROCCOLI_ENV` does not get set. This maintains the existing behavior of `broccoli build`, that `BROCCOLI_ENV` would be undefined if you don't manually define it. Doing `--environment=production` is basically an alias for `BROCCOLI_ENV=production broccoli build`

test/cli_test.js
oligriffiths: Well this is what I am saying. Should `BROCCOLI_ENV` flow through to `options.env`, I don't know if it should, but I agree this is bizarre.
Either we allow `BROCCOLI_ENV` to proxy through to `options.env` and thus keep the values in-sync, or we say that it's a one way operation, that `--environment` will set `BROCCOLI_ENV` but `BROCCOLI_ENV` will NOT set `--environment/options.env`

oligriffiths:
On further reflection, I don't think we should do this.
If someone is using `BROCCOLI_ENV` right now, this will continue to work.
If you opt to use `--environment` then you're opting into the *new way*, and as such you should adopt the `function` syntax. Saying it's deprecated but continuing to support its usage is confusing.
Thanks for the reviews.

PR #394, opened 2019-03-08
`environment` => `env` in README example

jfdnc:
Code example references undefined key `environment`. Change to `env` to reflect the actual key on the `options` object.


oligriffiths:
Oops, good catch

PR #395, opened 2019-04-03
Output port in use error if 2 servers started on the same port

oligriffiths:
This has always irritated me, when starting a server when you always have one running, and you get this nasty error:
image

So I made it nicer:
image

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -29,14 +29,16 @@ exports.serve = function serve(watcher, host, port, _connect) {
   server.http = http.createServer(server.app);
   server.http.listen(parseInt(port, 10), host);
   server.http.on('error', error => {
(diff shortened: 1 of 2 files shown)


oligriffiths:
@stefanpenner yeah for sure. TBH we should look to use this server in ember-cli

rwjblue:
I think this is good (bad error messages suck), but we should make the default behavior the same as what ember-cli does with `ember s --port 0` (it finds an available port automatically for you).

oligriffiths:
@rwjblue good idea, I think that should be a separate PR though

stefanpenner:
Lgtm!

stefanpenner:
The future pr may always want to default to port 0

PR #397, opened 2019-04-18
[WIP] Side effect free spike 2

oligriffiths:
```js
const BroccoliPlugin = require('broccoli-plugin');
const { UnwatchedDir } = require('broccoli-source');

class SideEffectFree extends BroccoliPlugin {
  constructor(inputNodes, options) {
    options = options || {};
    options.sideEffectFree = true;
    super(inputNodes, options);
  }

  __broccoliGetInfo__(builderFeatures) {
    let nodeInfo = super.__broccoliGetInfo__(builderFeatures);
    nodeInfo.sideEffectFree = true;
    return nodeInfo;
  }
}

class ConditionalPlugin extends SideEffectFree {
  build() {
    console.log('ConditionalPlugin.build(). Called when source dirs change. Revises every other build. 🚀');

    if (this.hasBuilt) {
      this.hasBuilt = false;
      console.log('ConditionalPlugin not revised! SlowPlugin should NOT be built. 🚫');
    } else {
      console.log('ConditionalPlugin revised! SlowPlugin should now be built. 🥦');
      this.revise();
      this.hasBuilt = true;
    }
  }
};

class SlowPlugin extends SideEffectFree {
  build() {
    console.log('SlowPlugin.build(). This should only be invoked when ConditionalPlugin changes ⚠️');
    console.log('SlowPlugin.build(). This is a slow plugin, delays by 2s 😴️');
    return new Promise(resolve => {
      setTimeout(() => {
        console.log('SlowPlugin.build(). Work finished, revising 🎉');
        this.revise();
        resolve();
      }, 2000);
    })
  }
};

class UnwatchedPlugin extends SideEffectFree {
  build() {
    console.log('UnwatchedPlugin.build(). This should only be called once as source dir is unwatched.');
    this.revise();
  }
};

class TestPlugin extends SideEffectFree {
  build() {
    console.log('TestPlugin.build(). Called only if SlowPlugin or BazPlugin revised. 👏');
    this.revise();
  }
};

class NormalPlugin extends BroccoliPlugin {
  build() {
    console.log('NormalPlugin.build(). This should always build as sideEffectFree = false');
  }
}

export default () => {
  const conditionalPlugin = new ConditionalPlugin(['app']);
  const slowPlugin = new SlowPlugin([conditionalPlugin]);
  const unwatchedPlugin = new UnwatchedPlugin([new UnwatchedDir('public')]);
  const testPlugin = new TestPlugin([slowPlugin, unwatchedPlugin]);
  return new NormalPlugin([testPlugin]);
};
```

3 separate builds (2 rebuilds)
```
Serving on http://localhost:4200

ConditionalPlugin.build(). Called when source dirs change. Revises every other build. 🚀
ConditionalPlugin revised! SlowPlugin should now be built. 🥦
SlowPlugin.build(). This should only be invoked when ConditionalPlugin changes ⚠️
SlowPlugin.build(). This is a slow plugin, delays by 2s 😴️
SlowPlugin.build(). Work finished, revising 🎉
UnwatchedPlugin.build(). This should only be called once as source dir is unwatched.
TestPlugin.build(). Called only if SlowPlugin or BazPlugin revised. 👏
NormalPlugin.build(). This should always build as sideEffectFree = false



Slowest Nodes (totalTime => 5% )              | Total (avg)
----------------------------------------------+---------------------
SlowPlugin (1)                                | 2005ms

Built - 2009 ms @ Mon Apr 29 2019 19:54:26 GMT-0400 (EDT)



ConditionalPlugin.build(). Called when source dirs change. Revises every other build. 🚀
ConditionalPlugin not revised! SlowPlugin should NOT be built. 🚫
NormalPlugin.build(). This should always build as sideEffectFree = false



Slowest Nodes (totalTime => 5% )              | Total (avg)
----------------------------------------------+---------------------
ConditionalPlugin (1)                         | 0ms
NormalPlugin (1)                              | 0ms
app (1)                                       | 0ms

Built - 1 ms @ Mon Apr 29 2019 19:54:29 GMT-0400 (EDT)
ConditionalPlugin.build(). Called when source dirs change. Revises every other build. 🚀
ConditionalPlugin revised! SlowPlugin should now be built. 🥦
SlowPlugin.build(). This should only be invoked when ConditionalPlugin changes ⚠️
SlowPlugin.build(). This is a slow plugin, delays by 2s 😴️
SlowPlugin.build(). Work finished, revising 🎉
TestPlugin.build(). Called only if SlowPlugin or BazPlugin revised. 👏
NormalPlugin.build(). This should always build as sideEffectFree = false



Slowest Nodes (totalTime => 5% )              | Total (avg)
----------------------------------------------+---------------------
SlowPlugin (1)                                | 2001ms

Built - 2003 ms @ Mon Apr 29 2019 19:54:34 GMT-0400 (EDT)
```


thoov:
@oligriffiths Some thoughts on this spike:

1. There is duplication of some of the revision tracking state (some of it lives in the source wrapper and some lives in the directory wrapper). I think this might make the mental model more complex than it needs to be.
2. Hanging `revised` off of the output facade `this.output.revise();` might not be the best place for 2 reasons. The first is that it requires us to commit to the output facade now (which I feel would need several rounds of though), and second is that say a plugin wants to create a file outside of something like `this.output.fs.writeFile()` it will need a way to notify broccoli that it has been revised which `this.revised()` might just be "simpler".
3. Relying on a revised boolean + something down the line to reset / settle that state feels like it could lead potential open the doors for some hard to track down bugs.

I feel like our 2 approaches are very similar with basically 1 things being different: who controls the "should build" logic. Does a wrapper ask upwards on its inputs for changes or does the wrapper itself maintain the current snapshot of revision counts of its inputs?

Since we are fairly close with each other we should discuss the point above at our next Wednesday meeting (cc: @stefanpenner) and come up with a vote. I think the best step forward is to choose one of these approaches and start "productionizing" them with tests and start working them into master.

oligriffiths:
@thoov in response:

1. Yeah this is a difficult one, it has to do with the order of operations.
  a. So first thing that happens is a file is changed
  b. this invokes the callback in the watcher
  c. the watcher has access to the source node wrapper, on which it calls revise()
  d. the watcher notifies the builder, which iterates each node
  e. each node needs its state reset from the previous build either before or after the build completes, this is currently done before build() is called on each node wrapper, but it may need to be done after
  f. as a result of this, we can't settle the source nodes directory revision state else we would wipe out the recently changed revision, hence the internal _revised.

I think the solution to the above is to reset/settle each node wrapper after the build completes, rather than before each build, this would eliminate the need for the internal revised tracking.

2. yes I think you're right, not exposing that interface right now is probably best
3. see point 1

thoov:
@oligriffiths A case that I didn't think about until talking w/ @stefanpenner was what if a change event fires during the build? Because this approach only uses a "boolean" it will "double set" the _revised value on the source nodes to true and the wrapper will lose context (when the reset happens) that it actually is still "dirty".

I believe we could add additional logic to guard against this case however it really feels like the state management approach in this is too complex for my liking (the need for a "reset" to happen afterwards).

The intrinsic simplicity in my approach I feel might be the better primitive. I feel that the output facade we will build in the future can be satisfied with both approaches it seems like we should go with the simpler design.

@oligriffiths What are your thoughts on this since this is different than what we came to on Wednesday?

oligriffiths:
I’m not super familiar with what the process is when a file is changed mid-build so I can’t really answer the implications for either of our approaches right now. 

I do know that we want to enable the ability to cancel a build mid way through when a change is detected, which might solve the issue here. 

Either way these apis are going to be internal so it’s kind of academic. 

Fortunately I have rwjblie sitting across the table from me so let me bend his ear and see what he thinks.

oligriffiths:
@thoov Ok so I chatted with rwjblue about this. He suggested we use a hybrid approach that the NodeWrapper should be the only place that knows about its revision state, but the decision of whether it needs to build should be computed by comparing the revision counter of self against the highest revision of the parents. If any parent has a revision > mine then we know we need to build. I've added in a sort of debounce so that the revision counter can only be incremented once per build, and it is "settled" after the build has resolved. Seems to work, doesn't require any updates to the builder.build() which is good and means that the state of each node is settled after its built.

Thoughts?

thoov:
@oligriffiths:

> Ok so I chatted with rwjblue about this. He suggested we use a hybrid approach that the 
> NodeWrapper should be the only place that knows about its revision state, but the decision of 
> whether it needs to build should be computed by comparing the revision counter of self against the > highest revision of the parents. If any parent has a revision > mine then we know we need to build.

I 100% agree with this.

> I've added in a sort of debounce so that the revision counter can only be incremented once per 
> build, and it is "settled" after the build has resolved. Seems to work, doesn't require any updates to > the builder.build() which is good and means that the state of each node is settled after its built.

I disagree with this because I feel like we lose the main benefit of a pure incremental counter (that we can increment anytime and we only care about revision delta's) and are basically just recreating a boolean.

The prime example would be this:

1. The initial build sets all NodeWrappers to `_revision` to 0.
2. A change fired from watcher_adapter (sane) simply increments that value.
3. This starts the rebuild flow and the above build checking logic is run.
4. If another change event is fired while this build is still in progress we can safely increment the `_revision` counter and just rely on broccolis built in rebuild logic to correctly handle this delta check. If we debounce the incrementing of `_revision` to only once pure build then the NodeWrappers will be in an incorrect state (their `_revision` will be 1 instead of 2 and then the rebuild happens the will not rebuild even though they changed).

> revision counter can only be incremented once per build

I think the biggest take away is that `_revision` does not translate to # of builds invoked but instead # of times sane was called on a node. This allows the NodeWrappers to know if they are "dirty" regards less of how many times we call build on them.

I would love to get @stefanpenner and @rwjblue to chime in more as I know they have great insights into our tradeoffs.

oligriffiths:
Rob mentioned we should look into this https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/reference/lib/tags.ts https://github.com/glimmerjs/glimmer-vm/blob/master/guides/05-validators.md#entity-tags

oligriffiths:
@thoov ok, I removed the debounce, this requires a settle at the end of the build which I think is fine.

thoov:
@oligriffiths Could you explain the value of calling settled after the build vs right here: https://github.com/broccolijs/broccoli/pull/397/files#diff-821cf385bc5c73e2f8c61081a2e50c2eR43. I get uneasy about calling something like this all the way after the build has completed vs just immediately updating it once we know it to be "out of date".

oligriffiths:
@thoov let me explain with an example:

Take the following initial state after initial build:
* source - 1
* transform1 - 1
* transform2 - 1

Then a change in source detected:
* source -2
* transform1: 1<2, build, but does not revise()
* transform1 - 1
* transform2: 1 !< 1 (no build)
* transform2 - 1

This is what we expect to happen. 
Now before the next build, we must update transform1&2 to a revision state of 2 in order for the revision number comparison to work. 

Then a change in source detected:
* source -3
* transform1: 2<3, build, and revise()
* transform1 - 3
* transform2: 2 < 3 build, no revise()
* transform2 - 2

If we call settle at the end of the transform build method then the counter will always be increments, as if revise() has been called, for example:

* source -2
* transform1: 1<2, build, but does not revise()
* transform1.settle()
* transform1 - 2
* transform2: 1 < 2 (build)
* transform2.settle()
* transform2 - 2

Settle HAS to happen after the build as node wrappers are not nested and built independently so a settle after the build would always lead to the next node wrapper building. 

Make sense?

thoov:
@oligriffiths I screen recorded the issue to hopefully demonstrate what I am seeing: 

I have created a simple brofile that is just a plugin (that waits for 5 seconds) and a simple watchdir as an input. After the initial build I change the contents of the file twice (meaning there is the initial change #1, and then #2 and #3) and using the settled approach it will not capture all 3 changes (only 2). This is because settling the state after the build counter "overrides" change #3 and doesnt call the build method:

not-working

Below is my pr demoing this use case working:

working

I think I misspoke about:

> Ok so I chatted with rwjblue about this. He suggested we use a hybrid approach 
> that the NodeWrapper should be the only place that knows about its revision state, > but the decision of whether it needs to build should be computed by comparing the > revision counter of self against the > highest revision of the parents. If any 
> parent has a revision > mine then we know we need to build.

Where I said I 100% agree. I agree that NodeWrappers should be where the state is held and that the build methods should do a revision check but I think its ok for NodeWrapper to know about their parents state. This last part simplifies so much for us while also covering all of the use cases. It allows us to not have to settle after the build is complete while keeping the correct revision state at all points.

I would like to hear from you about what it would take for your solution to handle the above use case?

Brofile used in these examples:

```
const BroccoliPlugin = require('broccoli-plugin');
const fs = require('fs');

class SlowPlugin extends BroccoliPlugin {
  constructor(inputNodes) {
    super([inputNodes], { needsCache: false, persistentOutput: true, sideEffectFree: true });
  }

  build() {
    return new Promise((resolve, reject) => {
      console.log("Slow plugin build started");

      const input = fs.readFileSync(`${this.inputPaths[0]}/foo.js`);
      console.log("Foo.js content: " +  input);
      
      setTimeout(() => {
        this.revise();
        resolve();
        console.log("Slow plugin build finished");
      }, 5000);
    });
  }

  __broccoliGetInfo__(builderFeatures) {
    let nodeInfo = super.__broccoliGetInfo__(builderFeatures);
    nodeInfo.sideEffectFree = true;
    return nodeInfo;
  }
};

export default () => {
  const slowPlugin = new SlowPlugin('src');
  return slowPlugin;
};
```

lib/builder.js
```diff
@@ -143,10 +143,15 @@ module.exports = class Builder extends EventEmitter {
         }),
       () => {
         this._cancelationRequest = null;
+        this._settleNodes();
       }
     );
   }
 
+  _settleNodes() {
+    this.nodeWrappers.forEach(nw => nw.settle());
```
stefanpenner: If watchman triggers `revise` after the specific node is invoked, but before the build completes, this approach will loose that information, resulting in erroneous skipped plugin build steps.

The other PR does not require 2 phase marking or cleanup, being less error prone is an important detail.

oligriffiths:
@thoov Let's go with your approach, it's all internal APIs so let's get this implemented and go from there.

PR #398, opened 2019-04-28
Add more watcher events

oligriffiths:
In order to support the Broccoli watcher in Ember CLI https://github.com/ember-cli/ember-cli/pull/8614, we need more information about the change event and some other events in order to replicate the existing Ember CLI functionality.

I added a test for Watcher as there didn't seem to be one. Plz lmk if I've done anything wrong.

Will add some tests, just wanted to get this open for now.


lib/watcher.js
```diff
@@ -58,7 +58,9 @@ module.exports = class Watcher extends EventEmitter {
       logger.debug('change', 'ignored: rebuild scheduled already');
       return;
     }
-    logger.debug('change');
+    logger.debug('change', event, filePath, root);
+    this.emit('change', event, filePath, root);
```
thoov: If I remember correctly the "change" event that ember-cli is expecting is a "results" object that contains: `directory`, `filepath`, `graph`, `outputChanges`, `totaltime`. Is the plan to modify that side to include just the above properties or should we emit the ones that ember-cli is expecting?

oligriffiths: Correct, and tbh I think that API isn’t good. A “change” event in ember-cli-sane-watcher indicates a build completed, as it does its own change tracking for individual files but doesn’t expose those events (add,change,delete). 

I am proposing we expose the change events and separate them from the build events. You can see the changes here https://github.com/ember-cli/ember-cli/issues/8614

oligriffiths:
@stefanpenner @thoov added tests, this should be good to review now

oligriffiths:
@stefanpenner added a test for Watcher as there doesn't seem to be one. Confused about a couple of things:

1. What is `_lifetimeDeferred` about?
2. If the build promise rejects with a failure, I don't see `_error` being called https://github.com/broccolijs/broccoli/blob/d152c60c3b7d7173e0f17ff3561207291d82ea44/lib/watcher.js#L51 I was debugging this test https://github.com/broccolijs/broccoli/blob/d152c60c3b7d7173e0f17ff3561207291d82ea44/test/watcher_test.js#L60 and asserting that `quitStart` was called, but it's not. Added some breakpoints and confirmed `_error` is not called. Is this right?

oligriffiths:
Ugh apvoyer failed, claims tests failed, dunno why...

thoov:
@oligriffiths Looks like the tests are failing for lint reasons. Could you fix that then LGTM!

oligriffiths:
Will fix. Heading to a flight rn tho

oligriffiths:
@thoov fixed tests. Ready to rock

oligriffiths:
👏

PR #401, opened 2019-05-17
Ensure tests run against all broccoli-plugin and broccoli-source versions

rwjblue:
These versions were released, but since the multidep.json file here was not updated we were not properly testing against them for cross-compatibility.


thoov:
@rwjblue 2.1.0 of broccoli-source was not published at the time of your last run. I just published it so could you trigger another run for this?

rwjblue:
Updated!

rwjblue:
Thanks for finishing this off @thoov!

PR #403, opened 2019-05-18
Pass watchedNodes to Watcher/WatcherAdapter

oligriffiths:
The intention of this PR is to make the `watchedNodes` an argument to the constructor of the watcher, as they're a required dependency and watcher can't work without them.
Additionally, NOT reaching into the builder to get them will help with the ember CLI integration of the watcher.

### DIFF
--- a/test/watcher_adapter_test.js
+++ b/test/watcher_adapter_test.js
@@ -120,19 +120,12 @@ describe('WatcherAdapter', function() {
     });
 
     it('throws if you try to watch a non array', function() {
-      expect(() => new WatcherAdapter()).to.throw(
-        TypeError,
-        `WatcherAdapter's first argument must be an array of SourceNodeWrapper nodes`
-      );
-
-      [null, undefined, NaN, {}, { length: 0 }, 'string', function() {}, Symbol('OMG')].forEach(
-        arg => {
-          expect(() => new WatcherAdapter(arg)).to.throw(
-            TypeError,
-            `WatcherAdapter's first argument must be an array of SourceNodeWrapper nodes`
-          );
-        }
-      );
+      [NaN, {}, { length: 0 }, 'string', function() {}, Symbol('OMG')].forEach(arg => {
+        expect(() => new WatcherAdapter(arg)).to.throw(
+          TypeError,
+          `WatcherAdapter's first argument must be an array of SourceNodeWrapper nodes`
+        );
+      });
     });
 
     it('throws if you try to watch a non node', function() {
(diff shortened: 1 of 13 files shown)


thoov:
@oligriffiths Should this be apart of the 3.0 release or should it come after?

oligriffiths:
@thoov I’d prefer if it were but I’m not going to be able to work much more on it until Monday

lifeart:
Just wondering, if it's a have sence - use current git revision for trigger rebuild if it changes (for upstream changes) or use requests from IDE like VSCode language server to get exact changed files (without watching overhead)

oligriffiths:
@lifesrt the fit revision doesn’t know about changed files, only committed ones. Not sure how this would help?

thoov:
@oligriffiths This looks good. Just to confirm that this is so we dont have to do this? (https://github.com/ember-cli/ember-cli/pull/8614/files#diff-9ff56008b40607a31bc04df7f41906cbR45)

test/watcher_adapter_test.js
```diff
@@ -88,11 +113,45 @@ describe('WatcherAdapter', function() {
...
+      expect(() => new WatcherAdapter([new TransformNodeWrapper()])).to.throw(
+        Error,
+        `[NodeWrapper:undefined undefined at undefined] is not a SourceNode`
+      );
+    });
+
+    it('throws if you try to watch a non-watchable node', function() {
+      expect(() => new WatcherAdapter([unwatchedNodeBasic])).to.throw(
+        Error,
+        `/Users/goli/Projects/broccoli/test/fixtures/basic it not watched`
```
stefanpenner: hard coded local path

oligriffiths: lol, oh yeah

oligriffiths:
@thoov correct. Also, the things to watch (nodes) are a direct dependency of this class. To me it makes more sense for them to be a mandatory argument rather than reaching in to the builder to get them

test/watcher_test.js
```diff
@@ -199,7 +202,7 @@ describe('Watcher', function() {
 
     it('does nothing if already quitting', function() {
       const adapterQuit = sinon.spy(adapter, 'quit');
-      const watcher = new Watcher(builder, { watcherAdapter: adapter });
+      const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
       watcher._quittingPromise = true;
```
stefanpenner: using private API in tests is an anti pattern, let's try and use public API's

oligriffiths: Sure thing

stefanpenner:
@lifeart I think asking git, for what is different, isn't a bad idea. We should explore that sometime.

oligriffiths:
@rwjblue valid points, will fix

PR #405, opened 2019-05-20
Require watchedNodes arguments to Watcher/WatcherAdapter

oligriffiths:
As the title says, per @rwjblue suggestion


stefanpenner:
LGTM

PR #408, opened 2019-05-28
[feature] stop current build, when rebuild is triggered

### LINKED ISSUE #407: stop current build, when rebuild is triggered.
When a funnel is being built, if a command to build again has been issued again, the funnel should stop its build and rebuild.

gabrielcsapo:
fixes #407 

_taken from @rwjblue_

```
Steps to repro:

Have a large build pipeline that takes a few seconds
change a file (kicks off a rebuild)
change another file during that rebuild
Expected result:

Broccoli should stop calling any node's .build method immediately after step 3 (essentially short circuiting the build) and start over
Actual result:

The build from step 2 completes then the build queued up in step 3 starts, therefore taking 2x the total time (time to complete two rebuilds vs time to do one full rebuild and whatever partial amount was previously completed when step 3 happens)
```

### DIFF
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -22,7 +22,7 @@ class Watcher extends EventEmitter {
   _ready: boolean;
   _quittingPromise: Promise<void> | null;
   _lifetime: {
-    promise?: Promise<void>,
+    promise?: Promise<void>;
     resolve?: (value: any) => void;
     reject?: (error: any) => void;
   } | null;
@@ -109,7 +109,7 @@ class Watcher extends EventEmitter {
   }
 
   async ready() {
-    return new Promise((resolve) => {
+    return new Promise(resolve => {
       this.on('ready', () => {
         resolve();
       });
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -109,11 +109,7 @@ class Watcher extends EventEmitter {
   }
 
   async ready() {
-    return new Promise(resolve => {
-      this.on('ready', () => {
-        resolve();
-      });
-    });
+    await new Promise(resolve => this.on('ready', resolve));
   }
 
   async _change(event: 'change', filePath: string, root: string) {
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -109,7 +109,7 @@ class Watcher extends EventEmitter {
   }
 
   async ready() {
-    await new Promise(resolve => this.on('ready', resolve));
+    await new Promise(resolve => this.once('ready', resolve));
   }
 
   async _change(event: 'change', filePath: string, root: string) {
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -95,6 +95,7 @@ class Watcher extends EventEmitter {
         try {
           await this.watcherAdapter.watch();
           logger.debug('ready');
+          this.emit('ready');
           this._ready = true;
         } catch (e) {
           this._error(e);
@@ -105,6 +106,14 @@ class Watcher extends EventEmitter {
     );
 
     return this._lifetime.promise;
+  }
+
+  async ready() {
+    return new Promise((resolve) => {
+      this.on('ready', () => {
+        resolve();
+      });
+    });
   }
 
   async _change(event: 'change', filePath: string, root: string) {
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -123,6 +123,9 @@ class Watcher extends EventEmitter {
     this._rebuildScheduled = true;
 
     try {
+      // cancel the current builder and wait for it to finish the current plugin
+      await this.builder.cancel();
+
       // Wait for current build, and ignore build failure
       await this.currentBuild;
     } catch (e) {
--- a/test/multidep.json
+++ b/test/multidep.json
@@ -1,6 +1,7 @@
 {
   "path": "test/multidep",
   "versions": {
+    "broccoli-merge-trees": ["4.1.0"],
     "broccoli-plugin": ["1.0.0", "1.1.0", "1.2.3", "1.3.0", "2.0.0", "2.1.0", "3.0.0"],
     "broccoli-source": ["1.1.0", "2.0.0", "2.1.1", "2.1.2", "3.0.0"]
   }
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -174,7 +174,7 @@ describe('Watcher', function() {
         }
 
         build() {
-          pluginsCalled['plugin1'] = pluginsCalled['plugin1'] ? pluginsCalled['plugin1'] + 1 : 1;
+    events.push('plugin1 build');
 
           return waitForPlugin1.promise;
         }
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -163,7 +163,7 @@ describe('Watcher', function() {
     });
 
     it('on change, rebuild is invoked and cancel is invoked from the build', function() {
-      const pluginsCalled = {};
+      const events = [];
 
       const waitForPlugin1 = defer();
       const waitForPlugin2 = defer();
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -170,7 +170,7 @@ describe('Watcher', function() {
 
       class Plugin1 extends Plugin {
         build() {
-    events.push('plugin1 build');
+          events.push('plugin1');
 
           return waitForPlugin1.promise;
         }
@@ -178,7 +178,7 @@ describe('Watcher', function() {
 
       class Plugin2 extends Plugin {
         build() {
-    events.push('plugin2 build');
+          events.push('plugin2');
 
           return waitForPlugin2.promise;
         }
@@ -202,18 +202,12 @@ describe('Watcher', function() {
           return changedBuild;
         })
         .then(() => {
-          expect(events).to.deep.equal([
-            'plugin 1',
-          ]);
+          expect(events).to.deep.equal(['plugin1']);
 
           return watcher.currentBuild;
         })
         .then(() => {
-          expect(events).to.deep.equal([
-            'plugin1',
-            'plugin1',
-            'plugin2',
-          ]);
+          expect(events).to.deep.equal(['plugin1', 'plugin1', 'plugin2']);
         });
     }).timeout(600000);
 
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -21,8 +21,16 @@ function defer() {
   return deferred;
 }
 
-function sleep(timeout = 10) {
-  return new Promise(resolve => setTimeout(resolve, timeout));
+function waitForWatcherToBeReady(watcher) {
+  return new Promise(resolve => {
+    if (watcher._ready === true) {
+      return resolve();
+    } else {
+      setTimeout(() => {
+        return resolve(waitForWatcherToBeReady(watcher));
+      }, 10);
+    }
+  });
 }
 
 const expect = chai.expect;
@@ -53,7 +61,7 @@ describe('Watcher', function() {
     async build(_, buildAnnotation) {
       return buildAnnotation;
     },
-    cancel() {
+    async cancel() {
       return Promise.resolve();
     },
   };
@@ -162,7 +170,7 @@ describe('Watcher', function() {
       expect(builderBuild).to.have.been.called;
     });
 
-    it('on change, rebuild is invoked and cancel is invoked from the build', function() {
+    it('on change, rebuild is invoked and cancel is invoked from the build', async function() {
       const events = [];
 
       const waitForPlugin1 = defer();
@@ -189,26 +197,23 @@ describe('Watcher', function() {
       const builder = new Builder(Merge([plugin1, plugin2]));
       const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
 
-      watcher._ready = true;
       watcher.start();
 
-      return sleep()
-        .then(() => {
-          const changedBuild = watcher._change('change', 'foo.js', 'root');
+      await waitForWatcherToBeReady(watcher);
+      {
+        const changedBuild = watcher._change('change', 'foo.js', 'root');
 
-          waitForPlugin1.resolve();
-          waitForPlugin2.resolve();
+        await waitForPlugin1.resolve();
+        await waitForPlugin2.resolve();
 
-          return changedBuild;
-        })
-        .then(() => {
-          expect(events).to.deep.equal(['plugin1']);
+        await changedBuild;
+      }
 
-          return watcher.currentBuild;
-        })
-        .then(() => {
-          expect(events).to.deep.equal(['plugin1', 'plugin1', 'plugin2']);
-        });
+      expect(events).to.deep.equal(['plugin1']);
+
+      await watcher.currentBuild;
+
+      expect(events).to.deep.equal(['plugin1', 'plugin1', 'plugin2']);
     }).timeout(600000);
 
     it('does nothing if not ready', function() {
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -178,7 +178,7 @@ describe('Watcher', function() {
 
       class Plugin2 extends Plugin {
         build() {
-          pluginsCalled['plugin2'] = pluginsCalled['plugin2'] ? pluginsCalled['plugin2'] + 1 : 1;
+    events.push('plugin2 build');
 
           return waitForPlugin2.promise;
         }
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -185,7 +185,7 @@ describe('Watcher', function() {
       }
 
       const plugin1 = new Plugin1([]);
-      const plugin2 = new Plugin2();
+      const plugin2 = new Plugin2([]);
       const builder = new Builder(Merge([plugin1, plugin2]));
       const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
 
--- a/test/watcher_test.js
+++ b/test/watcher_test.js
@@ -48,7 +48,7 @@ describe('Watcher', function() {
     async build(_, buildAnnotation) {
       return buildAnnotation;
     },
-    async cancel() { },
+    async cancel() {},
   };
 
   const adapter = {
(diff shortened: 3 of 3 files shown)


lib/watcher.ts
```diff
@@ -96,8 +96,8 @@ class Watcher extends EventEmitter {
 
     this._rebuildScheduled = true;
 
-    // Wait for current build, and ignore build failure
-    return Promise.resolve(this.currentBuild)
+    // If there is a current build, call cancel on it and ignore build failure
+    return Promise.resolve(this.builder.cancel())
```
mike-north: Are there cases where the cancellation needs to be communicated to plugins (i.e., if they have sidecar processes like a TS incremental compiler) so that they can take appropriate action? 

I'd like to understand context around important reasons to wait for `this.currentBuild` before proceeding with the next one.

gabrielcsapo: Hi @mike-north 👋 

>Are there cases where the cancellation needs to be communicated to plugins (i.e., if they have sidecar processes like a TS incremental compiler) so that they can take appropriate action?

Currently plugin creators can hook into the `cancel` hook to ensure that any existing work that needs to be cleaned up, can be cleaned up. 

Looking at the typescript compiler it looks like it doesn't have a cancel plugin. I can also see that the plugin is not returning a promise to build https://github.com/tildeio/broccoli-typescript-compiler/blob/master/src/plugin.ts#L51. 

Looking at the docs https://github.com/broccolijs/broccoli/blob/dfd95c07b7c1284f7a86945a1665c596345995a8/docs/node-api.md it says as a part of this that the build function must return a promise.

```
it must return a promise that is resolved on completion of the asynchronous work, or rejected if there is an error. Return values other than promises are ignored.
```

This code would not affect the typescript funnel as it is not cancellable from what I can infer.

> I'd like to understand context around important reasons to wait for this.currentBuild before proceeding with the next one.

This is primarily an issue for the project I am working on. Where you edit the code in multiple places and you must way (n) times for the build to finish. If a basic build takes 15 seconds and you edit 5 different places, you must wait 75 seconds. 

My main reason for adding this is why do we want for a build to succeed as that funnel is already waiting to build again. If we were to cancel the funnel, we would save time not having wait for stale builds to finish.

stefanpenner: @mike-north we don't current cancel running plugins, we simply let the current plugin complete but don't proceed to the next. This is already a reality plugins must be aware of as recoverable build errors will have the same behavior.

stefanpenner:
This is more or less how we ended cancellation to be used... thanks @gabrielcsapo for digging in.
I remember doing a bunch of prep work to enable this, I didn't realize I got so far that this was the only step remaining. I'm trying to recall if there was something else which prevented me from doing this last step.... It is very possible I simply ran out of time.

stefanpenner:
I __think__ the reason I didn't complete this at the time may have been related to ember-cli not yet using broccoli's built in watcher (which I believe it is now), it was still using the broccoli-sane-watcher. And I wanted a good end2end manual test in a real app.

@gabrielcsapo I'm assuming you have tested ^^ in your app at work? (That was the same app I wanted to test this with)

rwjblue:
Ya, I agree with @krisselden. The mocking in the tests here make it pretty hard to tell if the changes work in reality...

gabrielcsapo:
@rwjblue @krisselden I would be more than happy to fix, if you have time would love to peer and figure out how to make these tests more robust.

stefanpenner:
@gabrielcsapo I agree, mock's and spies are often anti-patterns in testing.

For cancellation the following pattern works: create a test with a multi step pipeline, give the test control to complete each step in the pipeline, cancel during 1 step, and assert the next step is not run. It should be relatively easy to adapt this pattern to trigger the cancelation via an input change.

gabrielcsapo:
@stefanpenner using your suggestion I have implemented a test case without using spies, mocks or stubs.

test/watcher_test.js
```diff
@@ -3,8 +3,28 @@ import chai from 'chai';
 import sinonChai from 'sinon-chai';
 import Sinon from 'sinon';
 import Watcher from '../lib/watcher';
+import broccoli from '..';
+const Builder = broccoli.Builder;
+const multidepRequire = require('multidep')('test/multidep.json');
 import SourceNodeWrapper from '../lib/wrappers/source-node';
 
+const Plugin = multidepRequire('broccoli-plugin', '1.3.0');
+const Merge = multidepRequire('broccoli-merge-trees', '4.1.0');
```
rwjblue: Why are we requiring these specific versions? Is it that these changes only work with those versions? If yes, then the change may be incompatible generally... 🤔 

We should probably be running these tests within a `multidepRequire.forEachVersion` loop...

test/watcher_test.js
```diff
@@ -3,8 +3,28 @@ import chai from 'chai';
...
+    deferred.resolve = resolve;
+    deferred.reject = reject;
+  });
+  deferred.promise = promise;
+  return deferred;
+}
+
+function sleep() {
+  return new Promise(resolve => setTimeout(resolve, 10));
+}
```
rwjblue: When I write a `sleep` function like this, I generally implement accepting a timeout value (defaulting if appropriate):

```suggestion
function sleep(timeout = 10) {
  return new Promise(resolve => setTimeout(resolve, timeout));
}
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+      const pluginsCalled = {};
+
+      const waitForPlugin1 = defer();
+      const waitForPlugin2 = defer();
+
+      class Plugin1 extends Plugin {
+        constructor(inputNodes, options) {
+          super(inputNodes || [], options);
+        }
+
```
rwjblue: Why do we need this? It seems to me like we control the nodes passed, and we could just make sure to pass the right thing when constructing.
```suggestion
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+          return waitForPlugin1.promise;
+        }
+      }
+
+      class Plugin2 extends Plugin {
+        constructor(inputNodes, options) {
+          super(inputNodes || [], options);
+        }
+
```
rwjblue: Same as above (RE: constructing with the right arguments):

```suggestion
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+      const waitForPlugin1 = defer();
+      const waitForPlugin2 = defer();
+
+      class Plugin1 extends Plugin {
+        constructor(inputNodes, options) {
+          super(inputNodes || [], options);
+        }
+
+        build() {
+          pluginsCalled['plugin1'] = pluginsCalled['plugin1'] ? pluginsCalled['plugin1'] + 1 : 1;
```
rwjblue: Once migrated to an array of "build events" (see prior comment), this can become:

```suggestion
    events.push('plugin1 build');
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+        }
+      }
+
+      class Plugin2 extends Plugin {
+        constructor(inputNodes, options) {
+          super(inputNodes || [], options);
+        }
+
+        build() {
+          pluginsCalled['plugin2'] = pluginsCalled['plugin2'] ? pluginsCalled['plugin2'] + 1 : 1;
```
rwjblue: Once migrated to an array of "build events" (see prior comment), this can become:

```suggestion
    events.push('plugin2 build');
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
       expect(builderBuild).to.have.been.called;
     });
 
+    it('on change, rebuild is invoked and cancel is invoked from the build', function() {
+      const pluginsCalled = {};
```
rwjblue: On second thought, maybe it would be better to make this an array:

```suggestion
      const events = [];
```

This allows us to do something like this (psuedo code):

```js
const events = [];

class Plugin1 extends Plugin {
  build() {
    events.push('plugin1 build');

    return waitForPlugin1.promise;
  }
}

class Plugin2 extends Plugin {
  build() {
    events.push('plugin2 build');

    return waitForPlugin2.promise;
  }
}

//...snip...
expect(events).to.deepEqual([
  'plugin1 build',
  'plugin2 build',
  'plugin1 build',
  //...etc
]);
```

Changing to this structure lets us assert not only on the exact count of `build` calls for each plugin, but their exact ordering.

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+        }
+
+        build() {
+          pluginsCalled['plugin2'] = pluginsCalled['plugin2'] ? pluginsCalled['plugin2'] + 1 : 1;
+
+          return waitForPlugin2.promise;
+        }
+      }
+
+      const plugin1 = new Plugin1();
```
rwjblue: Passing arguments to `PluginX` to avoid needing to customize `constructor` (and therefore simplifying the test a bit):

```suggestion
      const plugin1 = new Plugin1([]);
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+      const plugin2 = new Plugin2();
```
rwjblue: Passing arguments to `PluginX` to avoid needing to customize `constructor` (and therefore simplifying the test a bit):

```suggestion
      const plugin2 = new Plugin2([]);
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
       expect(builderBuild).to.have.been.called;
     });
 
+    it('on change, rebuild is invoked and cancel is invoked from the build', function() {
+      const pluginsCalled = {};
+
+      const waitForPlugin1 = defer();
+      const waitForPlugin2 = defer();
```
rwjblue: I don't understand why we need these deferred's at all. At the moment, they are both resolved at the same time. Can you explain what purpose they serve?

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+
+      const plugin1 = new Plugin1();
+      const plugin2 = new Plugin2();
+      const builder = new Builder(Merge([plugin1, plugin2]));
+      const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
+
+      watcher._ready = true;
+      watcher.start();
+
+      return sleep()
```
rwjblue: Why are we `sleep`ing? What are we waiting for here?

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+          const changedBuild = watcher._change('change', 'foo.js', 'root');
+
+          waitForPlugin1.resolve();
+          waitForPlugin2.resolve();
+
+          return changedBuild;
+        })
+        .then(() => {
+          expect(pluginsCalled['plugin1']).to.eq(1);
+          expect(pluginsCalled['plugin2']).to.eq(undefined);
```
rwjblue:
```suggestion
          expect(events).to.deep.equal([
            'plugin 1',
          ]);
```

test/watcher_test.js
```diff
@@ -139,6 +162,65 @@ describe('Watcher', function() {
...
+
+          return watcher.currentBuild;
+        })
+        .then(() => {
+          expect(pluginsCalled['plugin1']).to.eq(2);
+          expect(pluginsCalled['plugin2']).to.eq(1);
```
rwjblue:
```suggestion
          expect(events).to.deep.equal([
            'plugin1',
            'plugin1',
            'plugin2',
          ]);
```

test/watcher_test.js
gabrielcsapo: I just wanted to start a promise chain. I can just do Promise.resolve().then()

test/watcher_test.js
gabrielcsapo: It is simply to create dummy plugins and merge them into one output. There is probably a better way of doing this.

test/watcher_test.js
```diff
@@ -33,6 +53,9 @@ describe('Watcher', function() {
     async build(_, buildAnnotation) {
       return buildAnnotation;
     },
+    cancel() {
```
stefanpenner: nitpick: I would prefer `async cancel() { }`

test/watcher_test.js
stefanpenner: This new approach looks good to me, I think a small comment could likely address @rwjblue's concern.

It would seem we could rewrite the rest of this function as a chain of `awaits`

```js
await Promise.resolve() // wait for 1 turn of the micro-task queue
{
   const build = watcher._change('change', 'foo.js', 'root');

          waitForPlugin1.resolve();
          waitForPlugin2.resolve();

  await build;
} 

// ...
```

test/watcher_test.js
```diff
@@ -139,6 +162,55 @@ describe('Watcher', function() {
...
+          return waitForPlugin2.promise;
+        }
+      }
+
+      const plugin1 = new Plugin1([]);
+      const plugin2 = new Plugin2([]);
+      const builder = new Builder(Merge([plugin1, plugin2]));
+      const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
+
+      watcher._ready = true;
```
stefanpenner: can we avoid private API usage?

gabrielcsapo: removed this.

test/watcher_test.js
gabrielcsapo: I make this waitForTheWatcher to be ready. Basically it needs to boot up before we can do anything to it. We could run a full build for the watcher but then I would have to wire up the promises to only return on the second pass. This seems cleaner and easier to follow. Thoughts?

test/watcher_test.js
stefanpenner: @gabrielcsapo I think you should likely simplify this further, and decouple from the above helpers, and merger.

An example would look like

```js

// your deferred helper
function defer() {
  const result = {};
  result.promise = new Promise((resolve, reject) {
    result.resolve = resolve;
    result.reject =  reject;
  });
  return result;
}


class WaitingPlugin extends Plugin {
  constructor(inputNodes) {
      super(inputNodes);
      this._waiter = defer();
      this.buildCount = 0;
   }

   resolve(value) {
     this._waiter.resolve(value);
   }

   reject(reason) {
     this._waiter.resolve(reason);
   }

   async build() {
     try {
       await this._defer.promise;
     } finally {
       this.buildCount++;
       this._waiter = defer();
     }
   }
}

const first = new WaitingPlugin();
const second = new WaitingPlugin([first]);
const builder = new Builder(second);


```

   }

  
}
```

Comment (stefanpenner) [reply test/watcher_test.js]: ^^ would not only address robs concern, but also make it very easy to test different combinations if needed.

File: test/watcher_test.js [hunk 17]
```diff
@@ -3,8 +3,35 @@ import chai from 'chai';
...
+  deferred.promise = promise;
+  return deferred;
+}
+
+function waitForWatcherToBeReady(watcher) {
+  return new Promise(resolve => {
+    if (watcher._ready === true) {
+      return resolve();
+    } else {
+      setTimeout(() => {
```
stefanpenner: needing this timer makes me nervous, what exactly are we waiting on here?

test/watcher_test.js
```diff
@@ -3,8 +3,35 @@ import chai from 'chai';
...
+    deferred.resolve = resolve;
+    deferred.reject = reject;
+  });
+  deferred.promise = promise;
+  return deferred;
+}
+
+function waitForWatcherToBeReady(watcher) {
+  return new Promise(resolve => {
+    if (watcher._ready === true) {
```
stefanpenner: Using private API here isn't ideal. Maybe we should simply add this method to the watcher itself?

test/watcher_test.js
gabrielcsapo: I moved this to be within the watcher and no longer relies on setTimeout.

test/watcher_test.js
gabrielcsapo: Added it to the watcher itself.

test/multidep.json
```diff
@@ -1,6 +1,7 @@
 {
   "path": "test/multidep",
   "versions": {
+    "broccoli-merge-trees": ["4.1.0"],
```
stefanpenner: I believe this is no longer used

test/watcher_test.js
```diff
@@ -33,6 +48,9 @@ describe('Watcher', function() {
     async build(_, buildAnnotation) {
       return buildAnnotation;
     },
+    async cancel() {
+      return Promise.resolve();
```
stefanpenner: this line isn't needed, simply `async cancel() { }` is sufficient

test/watcher_test.js
```diff
@@ -139,6 +157,83 @@ describe('Watcher', function() {
...
+        }
+      }
+
+      const plugin = new TestPlugin([]);
+      const builder = new Builder(plugin);
+      const watcher = new Watcher(builder, [watchedNodeBasic], { watcherAdapter: adapter });
+
+      watcher.start();
+
+      watcher.ready().then(() => {
```
stefanpenner: We should avoid "dangling promises", this should be rewritten as `await watcher.ready()` or similar

stefanpenner:
released as v3.4.0 🎉

rwjblue:
FWIW, this is causing issues downstream (see https://github.com/ember-cli/ember-cli/issues/9127). I think we should probably revert until we have done an bit more investigation...

gabrielcsapo:
I will look into this, revert makes the most sense 😢

gabrielcsapo:
https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L169 so it looks like the ember-cli builder code catches any errors thrown by broccoli. Cancellation errors make sense as those should be ignored.

gabrielcsapo:
@stefanpenner do we need to throw on cancellation? or should this simply return?

```js
  async build() {
    if (this._cancelationRequest) {
      throw new BuilderError('Cannot start a build if one is already running');
    }
}
```

could be turned into 

```js
  async build() {
    if (this._cancelationRequest) { return; }
```

gabrielcsapo:
Looking at this code, should we throw when things are cancelled? Cancellations are expected especially in this case.

stefanpenner:
I'm working on a fix

PR #411, opened 2019-06-10
add console-ui

stefanpenner:
As per discussion with @oligriffiths  and @thoov. TL;DR, keeping this all in one place keeps maintenance down.

- [x] inline slow trees

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -137,7 +137,10 @@ module.exports = function broccoliCLI(args, ui = new UI()) {
           watcher.quit();
         }
       });
-      watcher.on('buildFailure', () => ui.writeLine('build failure', 'ERROR'));
+      watcher.on('buildFailure', err => {
+        ui.writeLine('build failure', 'ERROR');
+        ui.writeError(err);
+      });
 
       function cleanupAndExit() {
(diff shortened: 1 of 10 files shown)


lib/cli.js
```diff
@@ -66,7 +67,14 @@ module.exports = function broccoliCLI(args) {
         watcher.on('buildSuccess', () => outputTree.sync());
       }
 
-      actionPromise = broccoli.server.serve(watcher, options.host, parseInt(options.port, 10));
+      actionPromise = broccoli.server.serve(
+        watcher,
+        options.host,
```
thoov: Its starting to feel like maybe we should switch to a config argument instead of just the raw args:

```
broccoli.server.serve({
  watcher,
  host: options.host,
  port: parseInt(options.port, 10),
  ui
});
```

We don't have to do this as apart of this PR but it feels weird to have to pass multiple `undefined`'s

stefanpenner: yes, this would be a breaking change so we likely need to wait for v4.

oligriffiths:
This is good, thanks

stefanpenner:
released as 3.1.0

PR #415, opened 2019-06-27
Enable colored syntax highlighting in error messages from babel

astronomersiva:


stefanpenner:
nice

Two things:

* can you add a test?
* whats our plan to coordinate this capability with ember-cli-babel, so we can configure babel to provide ansi?

package.json
```diff
@@ -27,6 +27,7 @@
     "test:debug": "mocha --inspect-brk"
   },
   "dependencies": {
+    "ansi-html": "^0.0.7",
```
rwjblue: Eeck, 0.0.7 seems rough (though this package has ~ 4 million weekly downloads)

stefanpenner: ya, likely fine :P

astronomersiva: I borrowed it from webpack's [dev server](https://github.com/webpack/webpack-dev-server/blob/90d4a7c7ae4198b79ff03a67b0037e4bed3154e8/client-src/default/overlay.js#L6). There's a comment there that tells

> // The error overlay is inspired (and mostly copied) from Create React App (https://github.com/facebookincubator/create-react-app)
// They, in turn, got inspired by webpack-hot-middleware (https://github.com/glenjamin/webpack-hot-middleware).

😆

astronomersiva:
@stefanpenner,

> * whats our plan to coordinate this capability with ember-cli-babel, so we can configure babel to provide ansi coloring?

Since ember apps rely on broccoli-middleware instead of this middleware, we will have to make the same changes in its [error-handler](https://github.com/ember-cli/broccoli-middleware/blob/master/lib/utils/error-handler.js#L23).

We can then remove the `options.highlightCode = false` in ember-cli-babel.

This is how it looks after doing these steps.

Screen Shot 2019-06-27 at 21 13 02

We can probably change the cyan to something else to make it play along well with the page background.

Let me take a stab at writing a test for this.

test/server_test.js
```diff
@@ -106,6 +110,33 @@ describe('server', function() {
...
+      server.http.on('listening', resolve);
+      server.http.on('close', reject);
+      server.http.on('error', reject);
+    }).then(() => {
+      return got(`http://127.0.0.1:${PORT}/`).then(
+        () => {
+          throw new Error('should not be reached');
+        },
+        err => {
+          expect(err.body).to.include(`<span style="color:#ff00ff;">102</span>`);
```
astronomersiva: @stefanpenner, please check if this is fine. The current version of got in this repo doesn't support getting the body from errors. I had to update it to v9 to get it working. Hope that's fine.

stefanpenner: Looks like an excellent test to me.

stefanpenner:
@astronomersiva this LGTM, is the plan for ember-cli-babel to detect the version of broccoli installed an swap the boolean flag when the appropriate version is present or...

stefanpenner:
released as v3.1.2 🎉

astronomersiva:
@stefanpenner, yes that's what I have in mind as well. After porting these changes to broccoli-middleware, we will have to check for that version in ember-cli-babel to toggle the boolean like you said.

We should probably explore merging broccoli-middleware with the middleware in this repo as well.

stefanpenner:
@astronomersiva yes, unifying the middleware would be fantastic.

PR #417, opened 2019-07-13
Support serving over HTTPS

astronomersiva:
Basically an iteration of #296. Rebasing it turned out to be too hard so I handed up doing it from scratch.


lib/server.js
```diff
@@ -99,9 +125,12 @@ exports.serve = function serve(
...
   _connect = require('connect'),
   _process = process,
-  ui
+  ui,
+  ssl,
+  sslKey,
+  sslCert
 ) {
-  const server = new Server(watcher, host, port, _connect, ui);
+  const server = new Server(watcher, host, port, _connect, ui, ssl, sslKey, sslCert);
```
astronomersiva: This kind of looks error prone. We should consider changing this signature to accept an object instead of individual parameters like these.

astronomersiva:
a gentle reminder @stefanpenner

lib/server.js
stefanpenner: Sounds good, although I suspect we don't need to block this PR on it.

stefanpenner:
@astronomersiva thanks for the reminder, sorry about the delay!

PR #418, opened 2019-07-25
Use a more appropriate data structure for nodeWrappers.

stefanpenner:
Use a more appropriate data structure for nodeWrappers.

Since NodeWrappers lists:

* must be insertion ordered
* must contain only unique values (based on their key)
* support a fast existence / lookup operation

This PR switches it from an array to a map.

As broccoli pipelines grow large, simply searching the array to detect duplicates on insertion can become very costly, and switching to a Map addresses this specific bottleneck.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -20,6 +20,7 @@ const BuildError = require('./errors/build');
 const CancelationRequest = require('./cancelation-request');
 const Cancelation = require('./errors/cancelation');
 const logger = require('heimdalljs-logger')('broccoli:builder');
+const filterMap = require('./utils/filter-map');
 
 // Clean up left-over temporary directories on uncaught exception.
 tmp.setGracefulCleanup();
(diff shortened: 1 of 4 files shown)


stefanpenner:
I think we should also add heimdall instrumentation for https://github.com/broccolijs/broccoli/blob/master/lib/builder.js#L170

lib/builder.js
```diff
@@ -143,8 +143,11 @@ module.exports = class Builder extends EventEmitter {
           this.buildHeimdallTree(outputNodeWrapper);
         }),
       () => {
-        let buildsSkipped = this.nodeWrappers.filter(nw => nw.buildState.built === false).length;
-        logger.debug(`Total nodes skipped: ${buildsSkipped} out of ${this.nodeWrappers.length}`);
+        let buildsSkipped = filterMap(
```
rwjblue: Note for a future iteration: it would be really nice if heimdall-logger provided a way to _conditionally_ emit logging info. Right now we are forced to eagerly calculate the debugging output when in 99% of the time its never consumed / reviewed.

lib/builder.js
```diff
@@ -170,11 +173,8 @@ module.exports = class Builder extends EventEmitter {
   makeNodeWrapper(node, _stack) {
     if (_stack == null) _stack = [];
 
-    // Dedupe nodes reachable through multiple paths
-    for (let i = 0; i < this.nodeWrappers.length; i++) {
-      if (this.nodeWrappers[i].originalNode === node) {
-        return this.nodeWrappers[i];
-      }
+    if (this._nodeWrappers.has(node)) {
```
rwjblue: Just curious if you prefer this setup (using has + get) over using get directly and guarding for undefined results? 

e.g. this:

```js
let wrapper = this._nodeWrappers.get(node);

if (wrapper !== undefined) {
  return wrapper;
}
```

Since we never expect to _actually_ store `undefined` ^ would "work". I'm not questioning or asking you to change this, I'm just curious about your preference in general...

PR #431, opened 2019-09-11
[FEATURE] Adding input/output feature to Broccoli

SparshithNR:
This feature is focused to provide better control over the files Read/Write operations. Now input and output property in the plug-in will provide a proxy to the `fs` library which will allow us to modify the relative path passed to be combined with outputPath generated by broccoli. 

Consumer of broccoli plugins need not do `fs.writeFileSync(this.outputPath + '/' + 'out.md', 'text')`, rather directly call the `this.output.writeFileSync('out.md', 'text');`

Example:
```js
const Filter = require('broccoli-persistent-filter');

class TestFilter extends Filter {
    constructor(nodes) {
      super(nodes, {
        fsFacade: true
      });
    }

    processString(content) {
        return content.replace(/broccoli/gi, `filter`);
    }
    build() {
      // this.input will have access to all the fs operations except write operations
      let content = this.input.readFileSync('a.txt', 'utf8');
      content = this.processString(content);
      // this.output will have access to all the fs operations.
      this.output.writeFileSync('a.md', content);
    }
};
```

This is the first step towards informed writing of the outputPath. Once all the write operations go via `this.output`, we can make a decision whether there was any write operation happened in the build if not there is nothing to update going down the build tree. 

Corresponding changes are made in the following repos:

https://github.com/broccolijs/broccoli-node-info/pull/9
https://github.com/broccolijs/broccoli-node-api/pull/7

Needs the node-api changes before this commit. So that we won't get the ts compile-time error. 

Tasks
- [ ] bump version for  `broccoli-node-info`
- [ ] bump version for  `broccoli-node-api`


test/wrappers_test.js
```diff
@@ -146,3 +149,75 @@ describe('transform-node', function() {
     chai.expect(spy).to.have.been.calledWith();
   });
 });
+
+describe('output-wrapper', function() {
+  let output;
+
+  beforeEach(() => {
+    fs.mkdirSync('tmp');
```
rwjblue: Please avoid making tmp dirs in project, use something like the `tmp` package instead

test/wrappers_test.js
```diff
@@ -146,3 +149,75 @@ describe('transform-node', function() {
...
+
+  it('should throw if given location is empty/false', function() {
+    expect(() => output.writeFileSync(null, 'test')).to.throw('relativePath can not be null');
+  });
+
+  describe('should write to given location provided filemeta', function() {
+    it('getDestinationPath', function() {
+      let options = {
+        fileMeta: {
+          getDestinationPath(relativePath) {
```
rwjblue: This is a very bizarre API, why would you call `.writeFileSync` with an argument _different_ than where you wanted to write the file?!

SparshithNR: The thinking behind API was if a `broccoli-plugin` does a mergeTree/funnel just for renaming folder or filename, this api can help the plugin to avoid merging/funneling. `getDestinationPath` will provide a handler to test and change path and `prefix` will just prepend the path. It has a good use case in `broccoli-persistent-filter` where plugins do multiple merges and funneling before the final tree is sent to `broccoli-persistent-filter`. But not in general broccoli of use case.

lib/wrappers/output.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const mkdirp = require('mkdirp');
+const logger = require('heimdalljs-logger')('broccoli:outputWrapper');
+
+module.exports = class OutputWrapper {
```
thoov: Spoke with @SparshithNR offline and my vision for this was more about a proxy:

```
var callbackObject = {};
var fs = { foo: 10 };


var inputFacade = {
      get: function(target, propertyName) {
         if (target.hasOwnProperty(propertyName)) {           
           return fs[propertyName](this.outputPath, ...arguments);
         }
          
    }
}

var proxy = new Proxy(fs, inputFacade);
callbackObject.input.fs = proxy;
```

lib/wrappers/transform-node.ts
```diff
@@ -23,11 +23,9 @@ export default class TransformNodeWrapper extends NodeWrapper {
     });
 
     if (this.nodeInfo.inoutFacade) {
-      this.callbackObject = {
-        ...this.nodeInfo.getCallbackObject(),
-        input: new FSMerger(this.nodeInfo.inputNodes),
-        output: new OutputWrapper(this.outputPath)
-      };
+      this.callbackObject = this.nodeInfo.getCallbackObject();
```
stefanpenner: can we unit test the shape of this callbackObject? I want to ensure this API is locked down.

SparshithNR: I have added a test to cover this `if` branch. I forgot to push the changes. :)

lib/wrappers/output.ts
```diff
@@ -0,0 +1,61 @@
...
+  'rmdir',
+  'rmdirSync',
+  'mkdir',
+  'mkdirSync',
+  'createWriteStream',
+  'copyFile',
+  'copyFileSync'
+]);
+
+export default class OutputWrapper {
```
thoov: @SparshithNR  Would something like this work?

```js
const fs = require('fs');

const WHITE_LIST = new Set([
  'writeFileSync'
]);

export default function OutputWrapper(node) {
  return new Proxy(fs, {
    get(target, propertyName) {
      if (!WHITE_LIST.has(propertyName) || !target.hasOwnProperty(propertyName)) {
        throw new Error(`${propertyName} method is not allowed. Supported methods are: ${WHITE_LIST.join(', ')}`);
      }

      return () => {
        // check if the first arg is path like and warn?

        target[propertyName](node.outputPath, ...arguments);
      };
    }
  };
}
```

lib/wrappers/output.ts
```diff
@@ -0,0 +1,61 @@
+import { PathLike } from "fs";
+import TransformNodeWrapper from "./transform-node";
+
+const fsExtra = require('fs-extra');
+const fs = require('fs');
```
thoov: Duplicate import of fs

lib/wrappers/output.ts
```diff
@@ -0,0 +1,61 @@
...
+const logger = require('heimdalljs-logger')('broccoli:outputWrapper');
+const path = require('path');
+
+const WRITEOPERATION = new Set([
```
thoov: I think we should be more conservative on which methods we allow. I don't think we should allow the nonSync variants as that may conflict with the lifecycle of other builds. What are your thoughts on this? cc @stefanpenner

lib/wrappers/output.ts
SparshithNR: This will work. But I feel that, it will become hard to extend the features of OutputWrapper like, we want to add more than just a proxy to `fs` in output wrapper. For example, if we want to add one more proxy like `path`, we will not able to add it directly or if we want to add a revision or rebuild functions.
If we are not expecting this. Above snippet will work. :)

thoov:
Input API:

```
this.input.readFileSync('./relative-path');
this.input.existsSync('./relative-path');
this.input.lstatSync('./relative-path');
this.input.readdirSync('./relative-path');
this.input.statSync('./relative-path');
```

Expected
1. If absolution path an error is thrown
2. if unsupported method an error is thrown

Output API:

```js
this.output.readFileSync('./relative-path');
this.output.writeFileSync('./relative-path', 'data');
this.output.appendFileSync('./relative-path', 'data');
this.output.rmdirSync('./relative-path', {});
this.output.mkdirSync('./relative-path');
```

Expected:
1. Error thrown if absolute path given
2. if unsupported method an error is thrown

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
+import fs, { PathLike } from "fs";
+import TransformNodeWrapper from "./transform-node";
+import { isAbsolute } from 'path';
+
+const logger = require('heimdalljs-logger')('broccoli:outputWrapper');
+
+const WHITELISTEDOPERATION = new Set([
+  'readFileSync',
+'existsSync',
```
thoov: Can we fix the indentation here

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
...
+  'mkdirSync'
+]);
+
+export default function OutputWrapper (node: TransformNodeWrapper) {
+  return new Proxy(fs, {
+    get(target: any, propertyName: string): any {
+      return function() {
+        let [relativePath] = arguments;
+        if (isAbsolute(relativePath)) {
+          throw new Error(`Relative path is expected, path ${relativePath} is an absolute path. outputPath gets prefixed to the reltivePath provided.`);
```
thoov: `reltivePath` is misspelt. I would drop the second sentence.

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
...
+        }
+        let outputPath = node.outputPath + '/' + relativePath;
+        if(WHITELISTEDOPERATION.has(propertyName)) {
+          arguments[0] = outputPath;
+          logger.debug(`[operation:${propertyName}] at ${outputPath}`);
+          return target[propertyName](...arguments);
+        } else {
+          throw new Error(`Operation ${propertyName} is not whitelisted to use. Whietlisted operations are ${Array.from(WHITELISTEDOPERATION).toString()}`);
```
thoov: `Whietlisted` is misspelt. I would change it to: `Allowed operations are`

lib/wrappers/transform-node.ts
```diff
@@ -18,7 +20,15 @@ export default class TransformNodeWrapper extends NodeWrapper {
       outputPath: this.outputPath,
       cachePath: this.cachePath,
     });
-    this.callbackObject = this.nodeInfo.getCallbackObject();
+    //adding ts-ignore for now. Should be removed later when new version of broccoli-node-api is released
+    // @ts-ignore
+    if (this.nodeInfo.fsFacade) {
+      this.callbackObject = this.nodeInfo.getCallbackObject();
+      this.callbackObject.input = new FSMerger(this.nodeInfo.inputNodes).fs,
+      this.callbackObject.output = OutputWrapper(this);
```
rwjblue: It is odd to have a plain function start with an upper case letter that isn't invoked with `new`, maybe call this `buildOutputWrapper(this)`?

lib/wrappers/transform-node.ts
```diff
@@ -3,12 +3,14 @@ import rimraf from 'rimraf';
 import undefinedToNull from '../utils/undefined-to-null';
 import NodeWrapper from './node';
 import { TransformNodeInfo, CallbackObject } from 'broccoli-node-api';
+import OutputWrapper from './output';
 
+const FSMerger = require('fs-merger');
```
rwjblue: We probably need to increase bus factor in this library before leveraging. Need at least a couple of folks that can merge PR's and publish npm releases.

I've also opened up [a number of issues](https://github.com/SparshithNR/fs-merger/issues) RE: the implementation in fs-merger over on its repo.

lib/wrappers/transform-node.ts
```diff
@@ -18,7 +20,15 @@ export default class TransformNodeWrapper extends NodeWrapper {
...
     });
-    this.callbackObject = this.nodeInfo.getCallbackObject();
+    //adding ts-ignore for now. Should be removed later when new version of broccoli-node-api is released
+    // @ts-ignore
+    if (this.nodeInfo.fsFacade) {
+      this.callbackObject = this.nodeInfo.getCallbackObject();
+      this.callbackObject.input = new FSMerger(this.nodeInfo.inputNodes).fs,
+      this.callbackObject.output = OutputWrapper(this);
+    } else {
+      this.callbackObject = this.nodeInfo.getCallbackObject();
```
rwjblue: Why isn't this done unconditionally above this `if`? It seems like both sides need this...

lib/wrappers/transform-node.ts
```diff
@@ -18,7 +20,15 @@ export default class TransformNodeWrapper extends NodeWrapper {
       outputPath: this.outputPath,
       cachePath: this.cachePath,
     });
-    this.callbackObject = this.nodeInfo.getCallbackObject();
+    //adding ts-ignore for now. Should be removed later when new version of broccoli-node-api is released
+    // @ts-ignore
+    if (this.nodeInfo.fsFacade) {
+      this.callbackObject = this.nodeInfo.getCallbackObject();
+      this.callbackObject.input = new FSMerger(this.nodeInfo.inputNodes).fs,
```
rwjblue: Why are we mutating the callbackObject that the user-land plugin gave to us?

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default function OutputWrapper (node: TransformNodeWrapper) {
+  return new Proxy(fs, {
+    get(target: any, propertyName: string): any {
+      return function() {
+        let [relativePath] = arguments;
+        if (isAbsolute(relativePath)) {
+          throw new Error(`Relative path is expected, path ${relativePath} is an absolute path.`);
+        }
+        let outputPath = node.outputPath + '/' + relativePath;
```
rwjblue: What happens when `relativePath` is something like `../../../../etc/hosts`? I think it would "break" the scoping (basically the same issue as absolute paths).

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
...
+  'appendFileSync',
+  'rmdirSync',
+  'mkdirSync'
+]);
+
+export default function OutputWrapper (node: TransformNodeWrapper) {
+  return new Proxy(fs, {
+    get(target: any, propertyName: string): any {
+      return function() {
+        let [relativePath] = arguments;
```
rwjblue: Instead of destructuring here, you can put `relativePath` in the function definition:

```js
return function(relativePath) {
```

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
...
+  'writeFileSync',
+  'appendFileSync',
+  'rmdirSync',
+  'mkdirSync'
+]);
+
+export default function OutputWrapper (node: TransformNodeWrapper) {
+  return new Proxy(fs, {
+    get(target: any, propertyName: string): any {
+      return function() {
```
rwjblue: This will return a new function **every time _any_** fs method is called (which we expect to be quite a lot). We should probably avoid doing extra work repeatedly...

lib/wrappers/output.ts
```diff
@@ -0,0 +1,38 @@
+import fs, { PathLike } from "fs";
```
rwjblue: Where is `PathLike` used? Do we have linting for these files?

rwjblue:
I don't really see how this can practically be used. No broccoli plugin will be able to rely on the `this.input` and `this.output` properties existing, what do we expect them to actually do? Duplicate the implementations (branching on the presence of `this.input`/`this.output`)?

rwjblue:
It is also not clear (from the documentation or implementation) how we expect to actually leverage this code for something useful. I assume providing `this.input` and `this.output` in and of itself is not the goal, but instead that we do "something smart" when we know a plugin has been written to. Is that correct? Should the documentation say that?

lib/wrappers/transform-node.ts
SparshithNR: We do not have a method (not one that I know) to set these two properties into plugin other than mutating here.

SparshithNR:
> I don't really see how this can practically be used. No broccoli plugin will be able to rely on the `this.input` and `this.output` properties existing, what do we expect them to actually do? Duplicate the implementations (branching on the presence of `this.input`/`this.output`)?

This will be initially opt-in feature for all the plugins. So if they opt-in, they can use the `this.input` and `this.output`. I am going to add the changes to `broccoli-concat` and `broccoli-funnel` soon we merge this. 



> It is also not clear (from the documentation or implementation) how we expect to actually leverage this code for something useful. I assume providing `this.input` and `this.output` in and of itself is not the goal, but instead that we do "something smart" when we know a plugin has been written to. Is that correct? Should the documentation say that?
The goal is once all the broccoli plugins use `this.output` to write, we can mark for a rebuild rather than doing `mtime or hash check` to rebuild.

rwjblue:
> This will be initially opt-in feature for all the plugins. So if they opt-in, they can use the this.input and this.output. I am going to add the changes to broccoli-concat and broccoli-funnel soon we merge this.

Sure, that part is clear. The point I was making is that those broccoli plugins will have to keep around both sets of logic in order to function, since we do not properly "upgrade" them to always have `this.input`/`this.output`.

rwjblue:
> The goal is once all the broccoli plugins use this.output to write, when there is build in node didn't write anything, there is no need to build further down the build tree. We can just stop there.

Yep, I understand that also. However:

1) just because you use `this.output` to write some changes, doesn't mean you use it for **all** changes (and we do not currently document / suggest that you should)
2) it is not clear that this implementation allows that functionality
3) _adding_ the assumption mentioned here in my first point just above **after** releasing the `fsFacade` option would be a breaking change

SparshithNR:
> 1. just because you use `this.output` to write some changes, doesn't mean you use it for **all** changes (and we do not currently document / suggest that you should)
> 2. it is not clear that this implementation allows that functionality

Yes, at this moment we cannot assure that all the plugins will use this method. We may have to release a major version and `this.outputPath` shouldn't be allowed to access directly in the plugin. As of now, I thought this as a small step towards it. We will incrementally reach that point where all the writing will happen via `this.output`. 

> 3. _adding_ the assumption mentioned here in my first point just above **after** releasing the `fsFacade` option would be a breaking change

At this moment it is not a breaking change, in the future as I mentioned above it will be a breaking change.

rwjblue:
> We may have to release a major version and this.outputPath shouldn't be allowed to access directly in the plugin.

Right, my point is that we should hash out this interop issue and have a plan for it before landing this feature. In this case, a major version bump of Broccoli itself is not good enough, because the intent of the broccoli-node-info/broccoli-plugin APIs is **specifically** meant to make them work across major versions (and indeed you can see a number of broccoli plugins do test multiple broccoli versions 0.16,0.18,1.0.0, 2.0.0).

SparshithNR:
> Right, my point is that we should hash out this interop issue and have a plan for it before landing this feature. In this case, a major version bump of Broccoli itself is not good enough, because the intent of the broccoli-node-info/broccoli-plugin APIs is **specifically** meant to make them work across major versions (and indeed you can see a number of broccoli plugins do test multiple broccoli versions 0.16,0.18,1.0.0, 2.0.0).

We expect all the versions of the `broccoli` to work with all the versions of `broccoli-plugins`?

rwjblue:
Yep, that is exactly what I am saying.

See [the test configuration for that here](https://github.com/broccolijs/broccoli-plugin/blob/c12e78a1a62221b206ebcfb131909d897372eb6c/test/multidep.json#L4) and the latest test run (which includes Broccoli@2.0.0) [here](https://travis-ci.org/broccolijs/broccoli-plugin/jobs/596222598).

SparshithNR:
> Yep, that is exactly what I am saying.
> 
> See [the test configuration for that here](https://github.com/broccolijs/broccoli-plugin/blob/c12e78a1a62221b206ebcfb131909d897372eb6c/test/multidep.json#L4) and the latest test run (which includes Broccoli@2.0.0) [here](https://travis-ci.org/broccolijs/broccoli-plugin/jobs/596222598).

Tests break when I add the new broccoli-source version.

rwjblue:
Hmm, not sure what you mean. I just added Broccoli 2.1,2.2,2.3,3.0, and 3.1 in https://github.com/broccolijs/broccoli-plugin/pull/42. CI passed on the PR, and is passing on master.

Latest CI run (as of this moment) is green over at https://travis-ci.org/broccolijs/broccoli-plugin/jobs/596235974.

SparshithNR:
https://github.com/broccolijs/broccoli/pull/440

lib/wrappers/transform-node.ts
```diff
@@ -18,7 +17,13 @@ export default class TransformNodeWrapper extends NodeWrapper {
       outputPath: this.outputPath,
       cachePath: this.cachePath,
     });
+    //adding ts-ignore for now. Should be removed later when new version of broccoli-node-api is released
```
rwjblue: Can you make an issue to clean this up? I really would prefer to not have to keep these ts-ignores for too long...

lib/wrappers/transform-node.ts
```diff
@@ -18,7 +17,13 @@ export default class TransformNodeWrapper extends NodeWrapper {
...
     this.callbackObject = this.nodeInfo.getCallbackObject();
+    // @ts-ignore
+    if (this.nodeInfo.fsFacade) {
+      // @ts-ignore
+      this.nodeInfo.setFSFacade();
```
rwjblue: Should we pass the `input` and `output` arguments to this function?

SparshithNR: Changed it, now passes `input` and `output`

lib/wrappers/transform-node.ts
SparshithNR: I have tasks mentioned in the PR description. Once we are ready to Merge this PR. we will bump the `broccoli-node-api` and `broccoli-node-info` then merge this change.

rwjblue:
I think this gets closed now that we are going a different direction, right?

SparshithNR:
Closing as we are taking a different direction.

PR #432, opened 2019-09-12
Converting index, cli, and messages to Typescript

thoov:
Converts index, cli, and messages to Typescript

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -5,32 +5,42 @@ import path from 'path';
 import CliError from './errors/cli';
 import broccoli from './index';
 import messages from './messages';
+import ConsoleUI from '../types/console-ui';
 
 const promiseFinally = require('promise.prototype.finally');
 const WatchDetector = require('watch-detector');
 const UI = require('console-ui');
 
 interface ServeOptions {
-  prod?: boolean;
-  dev?: boolean;
-  host?: string;
+  host: string;
   port: string;
-  environment?: string;
-  outputPath?: string;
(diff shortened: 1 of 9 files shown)


lib/cli.ts
```diff
@@ -141,7 +171,7 @@ module.exports = function broccoliCLI(args, ui = new UI()) {
           watcher.quit();
         }
       });
-      watcher.on('buildFailure', err => {
+      watcher.on('buildFailure', (err: Error) => {
```
stefanpenner: is this actually error? I think we don't have the assurance, rather i believe it is `any`

lib/cli.ts
```diff
@@ -153,10 +183,10 @@ module.exports = function broccoliCLI(args, ui = new UI()) {
       process.on('SIGINT', cleanupAndExit);
       process.on('SIGTERM', cleanupAndExit);
 
-      actionPromise = promiseFinally(watcher.start().catch(err => ui.writeError(err)), () => {
+      actionPromise = promiseFinally(watcher.start().catch((err: Error) => ui.writeError(err)), () => {
```
stefanpenner: ^

lib/cli.ts
```diff
@@ -153,10 +183,10 @@ module.exports = function broccoliCLI(args, ui = new UI()) {
...
         builder.cleanup();
         process.exit(0);
-      }).catch(err => {
+      }).catch((err: Error) => {
```
stefanpenner: ^

PR #434, opened 2019-09-20
Convert watcher and watcher_adapter to typescript

thoov:
Also bumped some typescript related libraries

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,30 +1,23 @@
 export = {
   get Builder() {
-    debugger;
     return require('./builder');
   },
   get loadBrocfile() {
-    debugger;
     return require('./load_brocfile');
   },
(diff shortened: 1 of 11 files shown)


thoov:
cc: @stefanpenner @oligriffiths

lib/index.ts
```diff
@@ -9,15 +9,15 @@ export default {
     return require('./server');
   },
   get getMiddleware() {
-    return require('./middleware');
+    return require('./middleware').default;
```
stefanpenner: this looks like a breaking change, if our own code must use `.default` what if your consumers require?

thoov: Fixed, this was unintended. I also confirmed locally that what is being exported is the same

PR #435, opened 2019-09-26
Converting the remaining lib files to typescript

thoov:
Converting: load-brocfile, middleware, dummy-watcher, cancelation-request, server, and builder to typescript


lib/middleware.ts
```diff
@@ -141,7 +146,7 @@ module.exports = function getMiddleware(watcher, options) {
           response.writeHead(200);
           response.end(buffer);
         },
-        buildError => {
+        (buildError: BuildError) => {
```
stefanpenner: do we know for sure this is a build error and not an any?

PR #440, opened 2019-10-10
Adding new versions of broccoli-source & broccoli-plugin to multidep testing

SparshithNR:


rwjblue:
Sad to see this so out of date. When we release these related projects (`broccoli-source`, `broccoli-plugin`, `broccoli`, `broccoli-node-info`, etc) we need to ensure that the various projects are updated and testing properly against each other.

PR #443, opened 2020-02-18
Async cleanup

chriseppstein:
I'm on the hunt for rejection leaks in the broccoli code that cause the `UnhandledRejectionWarning` message to be displayed when there is a build error while running `ember serve`.

Since broccoli requires node 8, it can use async/await. So I've rewritten the builder's `build` method to use async/await instead of the complex promise chaining that it was doing before.

After I did this, one of the two unhandled rejection warnings I was seeing when I ran broccoli's tests appears to have gone away (I'm still seeing the warning in my app though).

I'm submitting this PR because I think the code is easier to read now.


lib/builder.ts
```diff
@@ -10,7 +10,6 @@ import BuildError from './errors/build';
 import CancelationRequest from './cancelation-request';
 import Cancelation from './errors/cancelation';
 import filterMap from './utils/filter-map';
-import promiseFinally from 'promise.prototype.finally';
```
stefanpenner: Removing this, and using async/await is a breaking change, since node 8 does not support `finally`.

I am fine with us releasing a new major that drops node 8, but it will require a longer release cycle to be available in ember-cli. 

If the issue can be fixed without making this change, it can be released as a patch version

lib/builder.ts
```diff
@@ -98,74 +104,67 @@ class Builder extends EventEmitter {
...
+  private async _buildAsync() {
     this.buildId++;
 
+    // Wipe all buildState objects at the beginning of the build
     for (const nw of this._nodeWrappers.values()) {
-      // Wipe all buildState objects at the beginning of the build
       nw.buildState = {};
+    }
+
+    await forceAsync(); // run the remaining code in the next tick.
```
stefanpenner: `await Promise.resolve()` + your comment should be sufficient.

lib/builder.ts
```diff
@@ -98,74 +104,67 @@ class Builder extends EventEmitter {
...
-            })
-            .then(() => nw.build()),
-          () => {
-            if (needsEndNode) {
-              this.emit('endNode', nw);
-            }
+    try {
+      // Build each node
+      for (const nw of this._nodeWrappers.values()) {
+        this._cancelationRequest && this._cancelationRequest.throwIfRequested()
```
stefanpenner: I believe this line should be contained within the next `try` so that the `Cancelation.isCancelationError(err)` bellow can catch it

lib/builder.ts
```diff
@@ -98,74 +104,67 @@ class Builder extends EventEmitter {
...
+        } catch (err) {
+           // Not clear whether .build() is supposed to be able to throw a
+           // cancellation error or not but the promise version allowed it.
+          if (Cancelation.isCancelationError(err)) {
+            throw err;
+          } else {
+            throw new BuildError(err, nw);
+          }
+        }
+        this._cancelationRequest && this._cancelationRequest.throwIfRequested()
```
stefanpenner: Lines 153 and 136 don't seem like they are in the right position. The previous algorithm does the following.

1. `this._cancelationRequest.throwIfRequested()`
2. `nw.build()` 
3. `this._cancelationRequest.throwIfRequested()`
4. re-throw cancellation error (from 1, 3), or wrap error in a build error for improved build related diagnostics.

package.json
```diff
@@ -93,5 +93,9 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "volta": {
+    "node": "12.16.0",
```
stefanpenner: usually we pin to node versions that are the lowest we technically support, this helps us catch missing features during development

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
+    "composite": true,
```
stefanpenner: TL;DR?

cc @mike-north

lib/builder.ts
```diff
@@ -28,6 +27,13 @@ interface BuilderOptions {
 
 type NodeWrappers = TransformNodeWrapper | SourceNodeWrapper;
 
+// This can be await'ed to force everything that follows to run in the next tick.
+function forceAsync(): Promise<void> {
```
stefanpenner: I don't think we need this, `await Promise.resolve()` + comment in the calling code should be sufficient.

lib/builder.ts
```diff
@@ -98,74 +104,67 @@ class Builder extends EventEmitter {
     }
   }
 
+  build() {
+    // return this.buildPromise();
+    if (this._cancelationRequest) {
+      return Promise.reject(new BuilderError('Cannot start a build if one is already running'));
```
stefanpenner: could you make `build` an async function, and then on this line we can `throw new BuildError(/* message */)`

lib/builder.ts
```diff
@@ -98,74 +104,67 @@ class Builder extends EventEmitter {
...
+    }
+    let promise = this._buildAsync();
+    this._cancelationRequest = new CancelationRequest(promise);
```
stefanpenner: I believe in the previous code, most failures within the now `buildAsync` would not cause the `this.cancellationRequest` from being re-set. Today, any failure in that could would prevent this.

Is that intentional?

lib/builder.ts
chriseppstein: It's not a breaking change. It's not being used anymore because with async/await try/finally works as expected in node 8. I was able to rewrite all the promises that were using `promiseFinally` to use await within a try/finally block instead.

lib/builder.ts
chriseppstein: I agree this is not a strict port of the promise code but I'm pretty certain these are functionally equivalent and I find my version to be more straightforward because it doesn't catch the cancellation error just to rethrow it.

stefanpenner:
fixed one async issue: https://github.com/broccolijs/broccoli/pull/445

tsconfig.json
chriseppstein: `composite` projects can be used with `tsc --build` for more efficient rebuilds (without `--build` all files are recompiled with every build). It also allows the project to be used as dev time compilation dependency which should make co-developement of broccoli with other broccoli-related projects when using `yarn link` if broccoli is set up as a project reference in that other project's `tsconfig.json` file.

It's a "best practice" to be a composite project.

lib/builder.ts
chriseppstein: fixed.

lib/builder.ts
chriseppstein: I thought that code might not be sufficient to force a next tick, but it seems to do the trick.

lib/builder.ts
chriseppstein: All that code does is catch it and re-throw it if it's a cancellation error. I don't see the point of moving this.

lib/builder.ts
chriseppstein: Using async/await is not a breaking change. The node support is already node 8+.

package.json
chriseppstein: Ok I downgraded this to node 8.

lib/builder.ts
chriseppstein: I find the way the is written now to be more straightforward. It runs the build. While the build is running, you can cancel it and you can't start a new built. I don't know why we want to allow exceptions to that policy for "most failures" that could occur within the buildAsync method.

chriseppstein:
> To ensure we do not allow this test suite to pass if unhandledRejection occurs, could we run the following code, in-process prior to the tests running?

You pushed this change in #445.

stefanpenner:
> > To ensure we do not allow this test suite to pass if unhandledRejection occurs, could we run the following code, in-process prior to the tests running?
> 
> You pushed this change in #445.

I actually need to re-think this idea, as the watcher holds promise, for a later request from the server to read. This can legitimately trigger the unhandled promise notification, although it isn't a real issue, in-fact it is "by design" only registered later and since the public API.

There are several approaches for us to take:

a.  create a faux promise that doesn't trigger the unhandled rejection warning; this would allow us to hide the warning, without breaking public API
b. have our watcher -> server promise, fulfill with a complex type`{ value: unknown, type: 'FULFILLED', 'REJECTED' }`; this allows the server full control, no warnings, but will also be a breaking change.
b. do something totally different; I don't really thing that is needed here.

lib/builder.ts
stefanpenner: You are correct using async/wait is not a breaking change, but returning a promise from `build()` that doesn't have `finally` is a breaking change. That being said, I just reading `promise.prototype.finally` code, and it does not return a promise with a `finally` method. 

So you are correct, this wont be a breaking change.

lib/builder.ts
stefanpenner: After investigation this further it appears this may cause some additionally unexpected behavior as if `this._cancelationRequest` is not present the build can also not be cancelled. So guarding the existence of `this._cancelationRequest` suggests there is now a period where the build cannot be cancelled.

stefanpenner:
Thank you for pointing out this code could be improved, the current approach makes the cancellation not work quite correctly, see: https://github.com/broccolijs/broccoli/pull/443#discussion_r385220733

closing in favor of a larger refactor: https://github.com/broccolijs/broccoli/pull/447

PR #447, opened 2020-02-27
Modernize

stefanpenner:
* use async functions when it makes sense
* add some comments
* fix some warnings and english
* tighten up some algorithms


lib/builder.ts
```diff
@@ -241,7 +232,7 @@ class Builder extends EventEmitter {
       }
     }
 
-    // For 'transform' nodes, recurse into the input nodes; for 'source' nodes,
+    // For 'transform' nodes, recurs into the input nodes; for 'source' nodes,
```
rwjblue:
```suggestion
    // For 'transform' nodes, recurse into the input nodes; for 'source' nodes,
```

PR #449, opened 2020-02-27
Upgrade typescript to 3.8 and fix compilation issue.

chriseppstein:
Note: typescript doesn't follow semver so I've locked it down to 3.8.x since there's no guarantee that 3.9 won't break the build like 3.8 did.

### DIFF
--- a/lib/builder.ts
+++ b/lib/builder.ts
@@ -309,7 +309,7 @@ class Builder extends EventEmitter {
     const tmpObj = tmp.dirSync({
       prefix: 'broccoli-',
       unsafeCleanup: true,
-      dir: this.tmpdir,
+      dir: this.tmpdir || undefined,
     });
 
     this.builderTmpDir = tmpObj.name;
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "source-map-support": "^0.5.16",
     "symlink-or-copy": "^1.2.0",
     "ts-node": "^8.0.3",
-    "typescript": "^3.6.3"
+    "typescript": "~3.8.0"
   },
   "engines": {
     "node": "8.* || >= 10.*"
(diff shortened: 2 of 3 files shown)


lib/builder.ts
```diff
@@ -309,7 +309,7 @@ class Builder extends EventEmitter {
     const tmpObj = tmp.dirSync({
       prefix: 'broccoli-',
       unsafeCleanup: true,
-      dir: this.tmpdir,
+      dir: this.tmpdir || undefined,
```
rwjblue: Why is this needed? Do we have types for `this.tmpdir` wrong?

package.json
```diff
@@ -89,7 +89,7 @@
     "source-map-support": "^0.5.16",
     "symlink-or-copy": "^1.2.0",
     "ts-node": "^8.0.3",
-    "typescript": "^3.6.3"
+    "typescript": "~3.8.0"
```
rwjblue: 👍👍

lib/builder.ts
chriseppstein: the `dir` option only accepts `string | undefined` but `this.tmpdir` is `string | undefined | null`.

This casts `null` to `undefined` when passing it to `dirSync`.

I would have changed the type of `this.tmpdir` to not allow null, but it's technically a breaking change because the property is public.

stefanpenner: The type of `tmp.dirSync`'s `dir` is overly strict, not allowing `null`.

stefanpenner: I can adjust differently, but yes the types are somewhat misaglined

`string | null | undefined`  cannot be assigned to `string | undefined`

chriseppstein: (I suppose this also casts `""` to `undefined`, but an empty string seems bad in this case)

stefanpenner: Actually, I think this is a better option as is. Due to both keeping broccolis public types table, and observing the type preferences of `tmp.dirSync`

PR #450, opened 2020-02-27
[BUGFIX] no longer cause spurious unhandled promise rejections warnings

stefanpenner:
This PR fixes broccoli to no longer unintentionally trigger node's `unhandledRejection` warnings.

Specifically these warnings where coming from `watcher.currentBuild` which is a public API, allowing external consumers to consider the latest builds outcome. For example the broccoli express middleware will on each request follow `watcher.currentBuild`. On fulfillment it will serve the file in question, or on rejection it will provide a useful error page.

`watcher.currentBuild` is the downstream promise of several internal watcher promises, each of which already handle rejections correctly appropriately.

This commit ensures external consumers of `currentBuild` are correctly notified of the final promise state, but without any external consumers rejections of `currentBuild` are considered handled.

---

The test here is the addition of failing the test if there are any unhandled rejection warnings.

### DIFF
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -5,7 +5,6 @@ import WatcherAdapter from './watcher_adapter';
 import SourceNodeWrapper from './wrappers/source-node';
 
 const logger = require('heimdalljs-logger')('broccoli:watcher');
-
 interface WatcherOptions {
   debounce?: number;
   watcherAdapter?: WatcherAdapter;
@@ -29,7 +28,7 @@ class Watcher extends EventEmitter {
   } | null;
 
   options: WatcherOptions;
-  currentBuild: null | Promise<void | null>;
+  _currentBuild: Promise<unknown>;
   watcherAdapter: WatcherAdapter;
   builder: any;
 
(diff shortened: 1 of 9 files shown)


lib/watcher.ts
```diff
@@ -39,18 +38,36 @@ class Watcher extends EventEmitter {
...
+    this._currentBuild = promise;
+
+    promise.catch(() => {
+      /**
+       * The watcher internally follows currentBuild, and outputs errors appropriately.
+       * Since watcher.currentBuild is public API, we must allow public follows
+       * to still be informed of rejections.
+       *
+       * This function accomplishes that.
+       */
```
hjdivad: I'm confused by this comment.  It looks to me that what it accomplishes is ensuring that a rejected `_currentBuild` is never an unhandled rejection.

The "this" that allowed public follows to handle rejections is that `_currentBuild === promise` rather than the chained promise from `catch`, but the comment reads to me that the "this" it's referring to is the catch handler itself.

```suggestion
      /**
       * The watcher internally follows currentBuild, and outputs errors appropriately.
       * Since watcher.currentBuild is public API, we must allow public follows
       * to still be informed of rejections.  However we do not want `_currentBuild` itself
       * to trigger unhandled rejections.
       *
       * By catching errors here, but returning `promise` instead of the chain from `promise.catch`,
       * both goals are accomplished.
       */
```

stefanpenner:
This appears to be related to https://github.com/ember-cli/ember-cli/issues/9007 I'll hold off on landing this until I fully understand that issue.

stefanpenner:
confirmed that this does not make ember-cli/ember-cli#9007 worse, rather it is important in landing the relevant fix.

PR #456, opened 2020-02-29
[Fixes #388] ensure broccoli-cli exits with the appropriate status co…

### LINKED ISSUE #388: Broccoli build returns 0 on build failure
When plugins fail throwing errors, `broccoli build` still returns exit code 0, despite the fact that it has not built successfully, and has not created the build target directory.

stefanpenner:
…de on build failure

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -86,7 +86,8 @@ export = function broccoliCLI(args: string[], ui = new UI()) {
         } catch (e) {
           if (e instanceof CliError) {
             ui.writeError(e.message);
-            return process.exit(1);
+            process.exitCode = 1;
+            return;
           }
 
           throw e;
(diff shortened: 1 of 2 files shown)


lib/cli.ts
```diff
@@ -173,6 +176,9 @@ export = function broccoliCLI(args: string[], ui = new UI()) {
       watcher.on('buildFailure', (err: any) => {
         ui.writeLine('build failure', 'ERROR');
         ui.writeError(err);
+        if (!options.watch) {
+          process.exitCode = 1;
```
stefanpenner: todo: test this scenario

peabnuts123:
Interesting. How does calling `process.exitz(1)` mean the process exits with the wrong code? I frequently do this in my own code. What's the benefit of setting `process.exitCode` first?

stefanpenner:
It doesn’t, it’s just a more cooperative way to shut down. Specifically the bug fix here was adding a missing exitCode assignment (see above inline comment)

peabnuts123:
Oh, I see! I had missed that there was an new exit scenario in there.

ryuran:
Something new for this PR @stefanpenner @rwjblue ?

stefanpenner:
released as v3.5.1 🎉

PR #457, opened 2020-02-29
[BUGFIX] Ensure build annotation is correctly passed to the builder, …

stefanpenner:
…and fulfilled via watcher.currentBuild.

This feature is used heavily by ember-cli, but was previously insufficiently tested here. My small refactoring, converting code from promise chaining to async await, inadvertently broke this.

### DIFF
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -54,7 +54,7 @@ class Watcher extends EventEmitter {
     return this._currentBuild;
   }
 
-  // TODO: this is an interim solution, pending a largerly cleanup of this class.
+  // TODO: this is an interim solution, pending a largely cleanup of this class.
   // Currently I can rely on understanding the various pieces of this class, to
   // know this is safe. This is not a good long term solution, but given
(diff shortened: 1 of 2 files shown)


lib/watcher.ts
```diff
@@ -90,17 +90,19 @@ class Watcher extends EventEmitter {
...
+          await this.watcherAdapter.watch();
+          logger.debug('ready');
+          this._ready = true;
+        } catch (e) {
+          this._error(e);
+        }
 
-      await this._build();
-    })());
+        return this._build();
```
stefanpenner: this line was the culprit

test/watcher_test.js
```diff
@@ -29,11 +29,9 @@ describe('Watcher', function() {
     watched: true,
   };
 
-  const buildResults = {};
-
   const builder = {
-    async build() {
-      return buildResults;
+    async build(_, buildAnnotation) {
```
stefanpenner: We should echo the buildAnnotation we where provided, that ensures this test suite can be useful

PR #460, opened 2020-04-16
[feature] add a buildFinished event to builder

gabrielcsapo:
# Motivation

I am currently working on a project that requires knowing when a builder has finished without blocking on waiting on the current build.

This allows me to subscribe to events just like the _beginNode_ and _endNode_ events.

### DIFF
--- a/lib/builder.ts
+++ b/lib/builder.ts
@@ -170,6 +170,8 @@ class Builder extends EventEmitter {
       await pipeline;
       this.buildHeimdallTree(this.outputNodeWrapper);
     } finally {
+      this.emit('buildFinished', pipeline);
+
       let buildsSkipped = filterMap(
         this._nodeWrappers.values(),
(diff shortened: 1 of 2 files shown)


rwjblue:
Isn't it possible to model this by chaining off of the existing promise? I don't _really_ want to expose both a promise based interface **and** an evented one if we can help it... :thinking:

gabrielcsapo:
@rwjblue this is wanting to listen to broccoli events without being the one to awaiting on the current build. How would I get the finished build if I didn’t want to block on waiting for the current build and knowing when the build started? Please see https://github.com/gabrielcsapo/broccoli-inspector/blob/master/server/index.js

gabrielcsapo:
```js
    watcher.on('buildSuccess', () => {
      socket.emit('buildFinished');
    });
```

this is exactly what I was looking for!

PR #463, opened 2020-04-17
[fix] running lint as a seperate CI task and fixes all errors to ensure CI passes

### LINKED ISSUE #437: Run linting in CI directly (not through mocha)

gabrielcsapo:
fixes #437 

# Motivation

To ensure that linting is run. The previous linting setup was not catching all issues before they were submitted.

I had to turn off `no-unused-vars` because it does not currently work with imported types. The issue here https://github.com/typescript-eslint/typescript-eslint/issues/1856.


.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
   rules: {
     'no-var': 'error',
     'no-console': 'off',
+    'no-unused-vars': 'off', // this is turned off because of the lack of scope analysis on imported types (https://github.com/typescript-eslint/typescript-eslint/issues/1856)
```
rwjblue, line 16: I think if you use their recommended config you don't have to explicitly disable this (they have their own variant of `no-unused-vars` that *is* type aware).

From their [readme](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin):

```
{
  "extends": ["plugin:@typescript-eslint/recommended"]
}
```

gabrielcsapo: I did this and fixed all the errors, I need to go add typing information to the imports, will do that hopefully tonight or tomorrow.

gabrielcsapo:
@rwjblue I found a bug in lib/cli.js it looks like we were passing a string but it was expecting an error instance. Which would silently fail and cause the ui to write undefined.

gabrielcsapo:
bump on merging this as it fixes an error with ui.writeError as well

gabrielcsapo:
bump on merging this as it fixes an error with ui.writeError as well

PR #470, opened 2020-06-21
Fix wrong package name in README

ursm:
I think broccoli-merge-trees is correct, as there is no npm package named broccoli-merge.


joliss:
Thanks!

PR #472, opened 2020-07-29
Exit process with code 1 when build fails

chiragpat:
**Why**
Currently when a build fails broccoli-cli exits with code 0. This causes a lot of different tools that rely on exit code to assume that build succeeded even though it in fact failed

**How**
- Added a flag buildFailed and toggled it if the catch block is invoked
- Swapped exit code to 1 in the finally block based on that flag

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -254,14 +254,20 @@ export = function broccoliCLI(args: string[], ui = new UI()) {
       process.on('SIGTERM', cleanupAndExit);
 
       actionPromise = (async () => {
+        let buildFailed = false;
         try {
           await watcher.start();
         } catch (e) {
+          buildFailed = true;
           ui.writeError(e);
         } finally {
           try {
             builder.cleanup();
-            process.exit(0);
+            if (buildFailed) {
+              process.exit(1);
+            } else {
+              process.exit(0);
+            }
           } catch (e) {
             ui.writeLine('Cleanup error:', 'ERROR');
             ui.writeError(e);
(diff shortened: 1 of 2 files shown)


stefanpenner:
https://github.com/broccolijs/broccoli/pull/456 requires a test be written for [one scenario](https://github.com/broccolijs/broccoli/pull/456#discussion_r386007317). Rather than landing this PR, could you help address that test? That way, we address all the issues in one step.

lib/cli.ts
```diff
@@ -254,14 +254,20 @@ export = function broccoliCLI(args: string[], ui = new UI()) {
       process.on('SIGTERM', cleanupAndExit);
 
       actionPromise = (async () => {
+        let buildFailed = false;
         try {
           await watcher.start();
         } catch (e) {
+          buildFailed = true;
```
rwjblue, line 261: Since it seems that the process is already going to die, can we do:

```suggestion
          process.exitCode = 1;
```

Instead (then we don't need either of the `process.exit` calls directly)..

chiragpat:
> #456 requires a test be written for [one scenario](https://github.com/broccolijs/broccoli/pull/456#discussion_r386007317). Rather than landing this PR, could you help address that test? That way, we address all the issues in one step.

Yup I missed that PR when I created this one. I will go ahead and add a test to that PR.

lib/cli.ts
chiragpat: Yup it makes sense. TIL process.exitCode.

chiragpat:
Closing this PR in favor of adding a test to #456

PR #474, opened 2020-12-02
add the ability to ignore absolute paths from watcher

ef4:
(This PR builds off https://github.com/broccolijs/broccoli/pull/473, for a smaller review see only https://github.com/broccolijs/broccoli/commit/5fda57e0645855fcd0186af9919af7e0caea8f26.)

This adds an option to mark certain absolute paths as ignored by the watcher.

The motivation here is that if you want to be able to rebuild arbitrary node packages (which can put their source code at the top level of their project if they want to), and those packages emit build artifacts into some of their own subdirectories (like `dist` as used by ember-cli), you end up watching your own output.

In this situation, it is straightforward to know which path(s) are your own output, and tell broccoli not to watch them.

`sane`'s `ignored` option is perfect for this, except that it's always relative to the watchedDir, so broccoli's Watcher must translate.

### DIFF
--- a/lib/watcher.ts
+++ b/lib/watcher.ts
@@ -11,6 +11,10 @@ interface WatcherOptions {
   debounce?: number;
   watcherAdapter?: WatcherAdapter;
   saneOptions?: sane.Options;
+
+  // list of absolute paths that we should not watch, even if they end up nested
+  // beneath dirs that we are watching.
+  ignored?: string[];
 }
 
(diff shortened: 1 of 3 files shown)


rwjblue:
FYI - Just landed https://github.com/broccolijs/broccoli/pull/473, doesn't seem like this needs a rebase though.

rwjblue:
Released in https://github.com/broccolijs/broccoli/releases/tag/v3.5.0

PR #476, opened 2021-02-11
Exit broccoli-cli with correct status code.

### LINKED ISSUE #388: Broccoli build returns 0 on build failure

chiragpat:
This extends PR #456 to add a missing test and one more catch block that was missing an assignment to process.exitCode

Fixes #388


stefanpenner (approved):
LGTM, thank you.

I played around with the resulting executable, and it appears to function correct.

stefanpenner:
released as v3.5.1 🎉

PR #478, opened 2021-04-29
Manually exit process

asakusuma:
Partial revert of https://github.com/broccolijs/broccoli/pull/476
to work around https://github.com/babel/broccoli-babel-transpiler/issues/169

Kept the tests and ensuring the correct exit code is called, but put back the manual process exiting.

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -157,8 +157,7 @@ export = function broccoliCLI(args: string[], ui = new UI()) {
         } catch (e) {
           if (e instanceof CliError) {
             ui.writeError(e);
-            process.exitCode = 1;
-            return;
+            return process.exit(1);
           }
 
           throw e;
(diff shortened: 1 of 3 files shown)


asakusuma:
The one failing job seemed to hit an unrelated tooling error:
```
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/rxjs/-/rxjs-6.5.2.tgz: ESOCKETTIMEDOUT".
info If you think this is a bug, please open a bug report with the information provided in "/Users/runner/work/broccoli/broccoli/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```

rwjblue:
Re-ran CI (yarn install failed on macOS)

stefanpenner:
This is a workaround, could an inline comment be added describing the rationale?

The proper solution is a little more involved, and requires adding a destructor concept to broccoli pipelines, and plugins.

stefanpenner:
opened an issue: https://github.com/broccolijs/broccoli/issues/479

Issue #485, opened 2021-08-05
feature request: allow promise as entrypoint for Brocfile.js


lifeart:
use case: we need to request some side things to resolve build files and folders, and it will be nice to have this ability before build, not inside broccoli plugin with side-effect

PR #486, opened 2021-08-05
feat: async function support as entrypoint for Brocfile

lifeart:
resolves: https://github.com/broccolijs/broccoli/issues/485

allow us to do something like this:

```js
const funnel = require('broccoli-funnel');

module.exports = async function() {
   const files = await sideApiToResolveNeededFiles('app');

   return funnel('app', { files } );
}

```

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -9,6 +9,17 @@ import ConsoleUI from '../types/console-ui';
 import WatchDetector from 'watch-detector';
 import UI from 'console-ui';
 
+enum EnvironmentType {
+  PRODUCTION = 'production',
+  DEVELOPMENT = 'development',
+}
+
+enum WatcherType {
+  POLLING = 'polling',
+  WATCHMAN = 'watchman',
+  NODE = 'node',
+}
+
 interface ServeOptions {
   host: string;
   port: string;
@@ -19,8 +30,8 @@ interface ServeOptions {
   outputPath?: string;
   cwd?: string;
   noWatch?: boolean;
-  watcher?: string;
-  environment: string;
+  watcher?:  WatcherType,
+  environment: EnvironmentType;
   prod?: boolean;
   dev?: boolean;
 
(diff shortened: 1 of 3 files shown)


lib/cli.ts
```diff
@@ -9,6 +9,17 @@ import ConsoleUI from '../types/console-ui';
 import WatchDetector from 'watch-detector';
 import UI from 'console-ui';
 
+enum EnvironmentType {
+  PRODUCTION = 'production',
+  DEVELOPMENT = 'development',
+}
+
+enum WatcherType {
```
stefanpenner, line 17: this seems good

lib/cli.ts
```diff
@@ -9,6 +9,17 @@ import ConsoleUI from '../types/console-ui';
 import WatchDetector from 'watch-detector';
 import UI from 'console-ui';
 
+enum EnvironmentType {
```
stefanpenner, line 12: I'm not sure if broccoli imposes any specific environment name, rather it's for the consumer to decide.
This may be better off as a string

lib/cli.ts
```diff
@@ -44,9 +55,10 @@ function buildBrocfileOptions(options: { environment: string }) {
   };
 }
 
-function getBuilder(options: { environment: string }) {
+async function getBuilder(options: { environment: string }) {
   const brocfile = broccoli.loadBrocfile(options);
-  return new broccoli.Builder(brocfile(buildBrocfileOptions(options)));
+  const instance = await Promise.resolve(brocfile(buildBrocfileOptions(options)));
```
stefanpenner, line 60: nice improvement!

stefanpenner (changes requested):

lib/cli.ts
lifeart: It's explicitly defined in cli options

image

lib/cli.ts
lifeart: it's predefined sane watcher options 

image

lib/cli.ts
lifeart: of course environments could be different (user-defined), but inside broccoli codebase we support only 2 options (explicitly), and we don't expose this type as part of public API

PR #490, opened 2021-09-24
feat: CORS headers for served assets

lifeart:
Use case:

* Static Java application with it's own serving logic and port (port: 5000)
* Broccoli server running in parallel, to simplify frontend development (port: 4200)

We point to broccoli-served assets from port 5000 to 4200 and getting CORS error.

This MR introduce CORS headers for static assets, and should solve this error.

### DIFF
--- a/lib/middleware.ts
+++ b/lib/middleware.ts
@@ -134,6 +134,9 @@ function handleRequest(
     return;
   }
 
+  response.setHeader('Access-Control-Allow-Origin', '*');
+  response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
+
   response.setHeader('Cache-Control', 'private, max-age=0, must-revalidate');
   response.setHeader('Content-Length', stat.size);
   response.setHeader('Content-Type', mime.contentType(path.extname(filename)));
--- a/test/server_test.js
+++ b/test/server_test.js
@@ -202,6 +202,8 @@ describe('server', function() {
       expect(body).to.eql('Hello');
       expect(headers['last-modified']).to.eql('Fri, 27 Jul 2018 17:23:02 GMT');
       expect(headers['cache-control']).to.eql('private, max-age=0, must-revalidate');
+      expect(headers['access-control-allow-origin']).to.eql('*');
+      expect(headers['access-control-allow-methods']).to.eql('GET, OPTIONS');
       expect(headers['content-length']).to.eql('5');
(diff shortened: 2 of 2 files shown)


lifeart:
partially related https://github.com/broccolijs/broccoli/pull/292, but i'm think it's simplier to land CORS, comparing to this stalled MR

lib/middleware.ts
```diff
@@ -134,6 +134,9 @@ function handleRequest(
     return;
   }
 
+  response.setHeader('Access-Control-Allow-Origin', '*');
+  response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
```
stefanpenner, line 138: I believe the original concern from @joliss was that if this was enabled by default, all broccoli development environments could leak information (anything files served by the development middleware) to malicious websites the developers navigates too while running a broccoli service. Although this may seem somewhat esoteric, it exposes a risk I do not feel comfortable with either.

Many security issues are the composition of several not super serious issues. For example, one very dangerous composition would be:

1) if broccoli development environments had `'Access-Control-Allow-Origin:': '*'`
2) the static file serving middleware, or a user provided middleware, had a path traversal vulnerability.
...
4) any file readable by the development server could be accessed (which compounds if users happen to run the broccoli process with `sudo`

Other scenarios would be well known files or endpoints that contain data someone malicious can predict, or fuz for.

--- 

The approach of allowing users to specify their own headers seems appropriate, as we provide a more secure default experience, and allow them to opt out if they want. We should also discourage `'Access-Control-Allow-Origin': '*'`, and encourage explicit development domains to guard against scenarios such as the one listed above

stefanpenner (changes requested):

lib/middleware.ts
stefanpenner: resurrecting #292 or something targeted specifically at CORS seems appropriate

lifeart: @stefanpenner could we have `Access-Control-Allow-Origin` for `localhost`, `0.0.0.0`, `192.168.0.1` ?

lifeart: or, we could add cli option `--allowed-origins` to support CORS

stefanpenner: Hosts that are aliases for localhost seems like a good compromise. Let me think about it through a touch more, and reach out to some other folks to confirm.

good idea!

PR #501, opened 2022-09-30
Update ansi-html to v0.0.8

### LINKED ISSUE #492: Can ansi-html dependency be shifted to ansi-html-community to address advisory?
ansi-html has a high priority advisory against it for uncontrolled resource consumption and has had no new versions since 2017 and no response from the author. ansi-html-community is a fork of it that issued a 0.0.8 release in early September to address the issue.  See [PR on the original repo](https://github.com/Tjatse/ansi-html/pull/20).  Can the dependency be shifted to the fork?

gorner:
Resolves #492.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@types/chai": "^4.2.9",
     "@types/chai-as-promised": "^7.1.2",
     "@types/express": "^4.17.2",
-    "ansi-html": "^0.0.7",
+    "ansi-html": "^0.0.8",
     "broccoli-node-info": "^2.1.0",
     "broccoli-slow-trees": "^3.0.1",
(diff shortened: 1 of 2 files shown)


Z-Zhao:
Please accept this pull request. My npm audit report looks very bad because of the ansi-html vulnerability.

gorner:
@Z-Zhao Unfortunately this project seems to have remained dormant as the Ember community has shifted their focus to building out Embroider as the eventual replacement for Broccoli.

For now I would recommend specifying the updated version as an override in your package.json. Assuming you're using NPM, this would be:

```json
"overrides": {
  "ansi-html": "^0.0.8"
}
```

(If you're using Yarn, use "resolutions" instead of "overrides".)

PR #506, opened 2024-04-25
Update sane to 5.0.1 to fix unset-value vulnerability

LucasHill:
There is an [vulnerability](https://security.snyk.io/vuln/SNYK-JS-UNSETVALUE-2400660) in the unset-value package. This is being brought in through sane 4.x. All ember projects depend on broccoli through ember-cli and other packages. Even though the security risks of this vulnerability are admittedly low, it is triggering automated security systems and would be an easy fix if we could release a patch. I do not see any API changes in sane 5.x but the broccoli tests will not run for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "mime-types": "^2.1.26",
     "resolve-path": "^1.4.0",
     "rimraf": "^3.0.2",
-    "sane": "^4.0.0",
+    "sane": "^5.0.1",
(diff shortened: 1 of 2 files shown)


mkszepp:
This PR fix also a vulnerability in `braces` v2.x as `sane` v4 brings this outdated version
https://github.com/advisories/GHSA-grv7-fg5c-xmjg

